PR #224, opened 2026-06-19
fix(worker): subscribe to per-node shell stream for node-targeted jobs (#3914)

sunapi386:
## Context

Companion to **aceteam-ai/aceteam#3916** for issue aceteam-ai/aceteam#3914.

Node-targeted MCP jobs (`terminal_exec`, `code_*`, file reads, node attachments, ping) were dispatched by the platform to a single shared per-org shell stream `jobs:v1:shell:org_{org}` that **every** online worker in the org consumes via the same consumer group. The first worker to read claimed the job regardless of the intended target node, so jobs ran on the wrong node or timed out. Confirmed live: a job targeted at node `1008` (ubuntu-gpu) was executed by a different node (`aceteamvm`).

## Root cause (worker side)

The worker only ever subscribed to the shared `jobs:v1:shell:org_{org}` stream — it had no way to claim *only* the jobs meant for it.

## Changes

Each worker now **additionally** subscribes to its own per-node stream:

```
jobs:v1:shell:org_{org_id}:node:{headscale_node_id}
```

keyed by the Headscale numeric node ID (`network.GetGlobalStatus().NodeID`), which is the same value the platform stores in `fabric_node_status.node_id` and targets on.

- **`cmd/work.go`** — new `nodeQueueName(org, nodeID)`; comment notes it MUST match the Python `build_node_queue` byte-for-byte. After the Headscale node ID resolves (post network reconnect, where it becomes available), the worker subscribes to its per-node stream via an optional `AddQueue` interface, type-switched over `*APISource` / `*RedisSource`. Skips cleanly when the node ID or org is unknown (the platform's consumer-presence fallback covers those nodes).
- **`internal/worker/api_source.go`** — `APISource.AddQueue(queue)`: appends a queue after construction (the Redis API proxy creates the consumer group lazily on first XREADGROUP). Ignores blanks/duplicates.
- **`internal/worker/redis_source.go`** — `RedisSource.AddQueue(ctx, queue)`: appends a queue and creates the consumer group + stream (`EnsureConsumerGroups` → MKSTREAM) so the platform dispatcher's consumer-presence check can see the stream immediately. Ignores blanks/duplicates.

`AddQueue` is intentionally **not** part of the `JobSource` interface — it's a Redis-only concept and the Nexus HTTP-polling source must not be forced to implement it.

### Why this is safe (rollout)

The platform dispatcher routes a targeted job to the per-node stream **only when that stream has a live consumer**, falling back to the shared stream otherwise. So:

- Upgraded worker (this PR) → consumer present on its per-node stream → targeted jobs delivered only to it.
- Older worker (no per-node subscription) → no consumer → dispatcher uses the shared stream → node still works as today.

Purely additive: the worker keeps consuming the shared shell stream for untargeted/any-node work. No behavior change for existing deployments until the platform side ships.

## Testing

```
go vet ./...        # clean
go test ./...       # all packages pass
go test ./cmd/ ./internal/worker/ -run 'AddQueue|NodeQueueName|ShellQueue' -v
```

| Test | Coverage |
|------|----------|
| `cmd.TestNodeQueueName` | cross-repo guarantee: the per-node stream name equals the Python `build_node_queue` string |
| `worker.TestRedisSourceAddQueueCreatesStreamAndDelivers` | `AddQueue` creates the per-node stream/group (MKSTREAM) and jobs enqueued there are delivered with the right `SourceQueue` |
| `worker.TestRedisSourceAddQueueIdempotentAndBlank` | blanks and duplicates ignored |
| `worker.TestAPISourceAddQueue` | API source appends, ignores blanks, dedupes |

**Note:** the platform dispatcher gates routing on `XINFO GROUPS` reporting `consumers > 0`. miniredis does **not** track XREADGROUP consumers (it reports 0 even after a real read), so that exact signal is verified against real Redis as a manual/integration item rather than in a unit test. Real Redis registers the consumer on the first `XREADGROUP`, which the worker issues every poll once the queue is added.

**Manual (post-deploy):** on a node connected to the AceTeam Network, run `citadel work --verbose` and confirm the log line `per-node shell stream: jobs:v1:shell:org_{org}:node:{id}` with `{id}` equal to the node's Headscale numeric ID, then dispatch a targeted `terminal_exec` from the platform and confirm it lands on this node.

## Related

- Platform PR: aceteam-ai/aceteam#3916
- Issue: aceteam-ai/aceteam#3914

### DIFF
--- a/cmd/work.go
+++ b/cmd/work.go
@@ -551,6 +551,45 @@ func runWork(cmd *cobra.Command, args []string) {
 		Debug("node meta set: node_id=%s, node_name=%s", nodeName, nodeName)
 	}
 
+	// Subscribe to this node's per-node shell stream so node-targeted MCP jobs
+	// (terminal_exec, code_*, file reads, node attachments) reach ONLY this node
+	// instead of being claimed by a greedy peer on the shared shell stream
+	// (issue #3914). Keyed by the Headscale numeric node ID, which matches the
+	// platform's fabric_node_status.node_id used for targeting. We still consume
+	// the shared shell stream for untargeted work, so this is purely additive and
+	// safe for older platform deployments that don't yet route per-node.
+	//
+	// AddQueue is a Redis-only concept, so it's accessed via an optional
+	// interface rather than the JobSource contract (the Nexus HTTP source does
+	// not implement it). Done before the run loop starts, so no locking needed.
+	if headscaleNodeID != "" {
+		perNodeOrgID := ""
+		if deviceConfig != nil {
+			perNodeOrgID = deviceConfig.OrgID
+		}
+		if perNodeOrgID == "" {
+			if manifest, _, mErr := findAndReadManifest(); mErr == nil && manifest != nil {
+				perNodeOrgID = manifest.Node.OrgID
+			}
+		}
+		if perNodeOrgID != "" {
+			perNodeQueue := nodeQueueName(perNodeOrgID, headscaleNodeID)
+			switch src := source.(type) {
+			case *worker.APISource:
+				src.AddQueue(perNodeQueue)
+			case *worker.RedisSource:
+				if err := src.AddQueue(ctx, perNodeQueue); err != nil {
+					fmt.Fprintf(os.Stderr, "   - Warning: per-node stream subscribe failed: %v\n", err)
+				}
+			}
+			Debug("per-node shell stream: %s", perNodeQueue)
+		} else {
+			Debug("per-node shell stream skipped: org id unknown")
+		}
+	} else {
+		Debug("per-node shell stream skipped: Headscale node ID unavailable")
+	}
+
 	// Open usage store for per-job compute tracking
 	var usageStore *usage.Store
 	if nodeDir, err := platform.DefaultNodeDir(""); err != nil {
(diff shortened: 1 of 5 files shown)


sunapi386:
## Live 2-node verification runbook (the miniredis gap)

Confirmed both target nodes are online on the mesh:
- **aceteamvm** — linux/amd64, Tailscale `100.64.0.14`
- **desktop-6ukhjan** — Windows 11, amd64, Tailscale `100.64.0.52`

Per-node shell stream name (from `cmd/work.go` `buildNodeQueue`, byte-identical to the Python `build_node_queue`):
```
jobs:v1:shell:org_{org_id}:node:{node_id}
```

### Prereqs
1. Build the #224 branch for both platforms from an **isolated** worktree (do not touch the shared checkout):
   `git worktree add /tmp/cit-3914 fix/3914-per-node-stream-subscription && cd /tmp/cit-3914 && ./build.sh --all`
2. Install the built binary on **aceteamvm** (linux/amd64) and **desktop-6ukhjan** (windows/amd64).
3. Run the aceteam #3916 dispatcher (per-node routing + fallback) against the same Redis the nodes use, so node-targeted jobs route to the per-node stream.

### Steps
1. Start `citadel work` on **both** nodes (same org, same Redis).
2. **Consumer presence (the part miniredis can't cover):** on the live Redis, for each node confirm the per-node group has a live consumer:
   `XINFO GROUPS jobs:v1:shell:org_{org}:node:{nodeId}` → assert `consumers >= 1`.
   This is what the dispatcher's `_queue_has_consumers` gate reads.
3. **Targeted delivery:** dispatch a SHELL_COMMAND to **aceteamvm only** (via `terminal_exec`/MCP, or `XADD` to its per-node stream). Assert aceteamvm executes it and **desktop-6ukhjan does not**. Repeat targeting the Windows node; assert the reverse.
4. **Backward-compat fallback:** with no per-node consumer (stop one worker), confirm a node-targeted job falls back to the org-level `jobs:v1:shell:org_{org}` stream and is still delivered (no lost jobs).

### Pass criteria
- `XINFO GROUPS` shows `consumers >= 1` on each node's per-node stream while its worker runs.
- A targeted job lands **only** on the intended node (verified on both linux→aceteamvm and windows→desktop-6ukhjan).
- Fallback delivers when the per-node consumer is absent.

Cross-ref: aceteam-ai/aceteam#3916 (dispatcher side). Keeping this PR draft until the above passes on the live fabric.

Issue #159, opened 2026-06-15
feat: persist node identity across live ISO reboots


sunapi386:
## Problem

Every time a Citadel OS live ISO reboots, `citadel init` registers a new Headscale node with a new ID (e.g., 913 → 950 → 952). The old nodes go stale in `fabric_node_status`, cluttering the fabric dashboard with multiple "debian" entries.

## Root Cause

Live ISOs have no persistent storage by default. The `~/.citadel-cli/config.yaml` (containing `device_api_token` and org_id) and Headscale node state are lost on every reboot.

## Solutions

1. **Persistence partition** — the ISO supports `persistence` boot param. If a `persistence` ext4 partition exists on the USB, the overlay filesystem persists `~/.citadel-cli/` across reboots. The `--with-persistence` build flag generates a setup script.

2. **Re-registration with same identity** — `citadel init` could check if a node with the same hostname + org already exists in Headscale and reclaim it instead of creating a new one. This requires API support.

3. **Auto-cleanup** — the platform should garbage-collect `fabric_node_status` rows that have been offline for >24h and were registered from a live ISO (detectable via hostname pattern or a `live_iso: true` flag in heartbeat metadata).

## Impact

- Fabric dashboard shows multiple stale "debian" nodes
- Each reboot consumes a new Headscale pre-auth key
- Node-targeted routing (#3752) won't work if the node_id keeps changing

---
*Generated by Claude*

PR #255, opened 2026-06-20
fix: prevent node ID churn on restart + log registration to file

### LINKED ISSUE #246: Node fails to come online on fabric after v2.38.0 restart (registers stale, churns ID)
## Symptom
After restarting the citadel agent on **v2.38.0**, the node does **not** come online on the fabric. `terminal_list_nodes` / `nexus_list_nodes` show the prior node id (1015) as **offline** and no fresh online registration for this machine, even though the local `citadel` foreground process is running (~minutes uptime). Only unrelated nodes (1007 Windows) show online.

## Context
- Repeated restarts this session minted NEW node ids each time (702 → 1008 → 1015 → now none), i.e. the identity churn tracked in #237.
- The agent's session output goes to the terminal; the file log (`~/.citadel-cli/logs/latest.log`) only captured `command: citadel version`, so the agent's connect/register sequence is not observable from disk (this is the visibility gap #236 addresses — but #236's `/agent` introspection needs the node online to answer).

## Why it matters
A node that silently fails to register is invisible to job routing (node-targeted terminal_exec / file attachments time out) and to the new #236 introspection. Combined with the identity churn, an operator cannot tell whether the node is up.

## Asks
1. Repro: on v2.38.0, does `citadel` (foreground) reliably register online within N seconds? Capture the connect/register/heartbeat sequence.
2. Ensure the agent logs its registration result (online + node id) to the **log file**, not just stdout (extends #234/#236 logging).
3. Check for a v2.38.0 regression in the startup connect path; rule out tsnet/Nexus auth-key staleness after restart.

Related: #236, #237, #234, aceteam #3924.

sunapi386:
## Context

Fixes #246. After restarting the citadel agent on v2.38.0+, nodes fail to come online reliably. The node registers stale, and repeated restarts mint NEW node IDs each time (702 -> 1008 -> 1015 -> none). The agent logs its connect sequence to stdout but NOT to the log file, making post-mortem diagnosis impossible from `~/.citadel-cli/logs/latest.log`.

## Summary

Three root causes identified and fixed:

**1. Identity churn -- stale node not reclaimed before fresh connect** (`cmd/reconnect.go`)
When `recoverStaleVPN()` falls back from IP-preserving reconnect to `ClearState()` + fresh connect, the old Headscale node entry was NOT reclaimed. The reclamation logic (`reclaimStaleNodeByHostname`) existed in `cmd/init.go` but was never called from the work command's recovery path. Each restart that triggered a fresh connect created a new Headscale node while old ones accumulated stale in the dashboard. Now calls `reclaimStaleNodeByHostname()` before `ClearState()`.

**2. Premature stale detection -- single 10s attempt before giving up** (`internal/network/singleton.go`)
`VerifyOrReconnect()` made a single no-authkey reconnect with a 10s timeout. On boot, the network interface is often not ready within 10s, causing a false "stale" verdict that cascaded into `ClearState()` and a new Headscale ID. Now retries 3 times with escalating backoff (0s, 10s, 15s) before declaring stale. Worst case ~55s, well under 2 minutes. Non-stale errors (permission denied, missing state dir) bail immediately without retry.

**3. Registration not logged to file** (`cmd/work.go`, `cmd/reconnect.go`)
Key VPN status messages used `fmt.Println`/`fmt.Printf` (stdout-only) instead of `Log()`. The `Log()` function always writes to the log file, but many startup messages bypassed it. All critical registration-path messages now use `Log()`. Added a post-resolution summary line: `registered: online=true node_name=X headscale_id=Y state_dir=Z`.

## Test plan

| Area | Tests | Coverage |
|------|-------|----------|
| Retry constants | `TestReconnectAttemptsConstant` | Bounds check: 2-5 attempts |
| Backoff timing | `TestReconnectRetryBackoffTiming` | Worst case under 2 min, over 20s |
| Recovery no-token | `TestRecoverStaleVPNNoToken` | nil config + empty token |
| Recovery fetch fail | `TestRecoverStaleVPNFetchAuthkeyFails` | Unreachable server returns authkey error |
| Stale error detection | `TestIsStaleStateError` (existing) | 11 error patterns |
| All existing tests | Full suite | `go test ./internal/network/... ./cmd/` passes |

```
$ go test ./internal/network/... ./cmd/ -count=1
ok  	github.com/aceteam-ai/citadel-cli/internal/network	0.250s
ok  	github.com/aceteam-ai/citadel-cli/cmd	0.339s
```

## Related

- Closes #246
- Uses `reclaimStaleNodeByHostname()` from `cmd/init.go` (issue #159)
- Uses `ReclaimStaleNode()` from `internal/nexus/reclaim.go`

### DIFF
--- a/cmd/reconnect.go
+++ b/cmd/reconnect.go
@@ -147,37 +147,45 @@ func attemptVPNRecovery(ctx context.Context, deviceConfig *DeviceConfig, hostnam
 // 'citadel reconnect'). Also used by the --force path to avoid
 // duplicating fetch+connect logic.
 func recoverStaleVPN(ctx context.Context, deviceConfig *DeviceConfig, hostname, apiBaseURL string) VPNRecoveryResult {
-	Debug("VPN state is stale, attempting recovery...")
+	Log("VPN state is stale, attempting recovery (state_dir=%s, has_state=%v)...",
+		network.GetStateDir(), network.HasState())
 
 	if deviceConfig == nil || deviceConfig.DeviceAPIToken == "" {
+		Log("no device API token available for auto-recovery")
 		return VPNRecoveryResult{
 			Err: fmt.Errorf("no device API token available for auto-recovery"),
 		}
 	}
 
 	// Fetch a fresh authkey from the platform
-	Debug("requesting fresh authkey from %s", apiBaseURL)
+	Log("requesting fresh authkey from %s", apiBaseURL)
 	freshKey, fetchErr := network.FetchFreshAuthkey(ctx, apiBaseURL, deviceConfig.DeviceAPIToken)
 	if fetchErr != nil {
-		Debug("failed to fetch fresh authkey: %v", fetchErr)
+		Log("failed to fetch fresh authkey: %v", fetchErr)
 		return VPNRecoveryResult{
 			Err: fmt.Errorf("could not fetch fresh authkey: %w", fetchErr),
 		}
 	}
 
 	// Attempt 1: reconnect with existing state + fresh key (preserves IP)
-	Debug("attempting reconnect with existing state (IP-preserving)...")
+	Log("attempting reconnect with existing state (IP-preserving)...")
 	if ok, reconnErr := network.ReconnectWithAuthKey(ctx, freshKey); reconnErr == nil && ok {
-		Debug("reconnected with existing state (IP preserved)")
+		Log("reconnected with existing state (IP preserved)")
 		return VPNRecoveryResult{Connected: true, IPPreserved: true}
 	} else {
-		Debug("IP-preserving reconnect failed: %v", reconnErr)
+		Log("IP-preserving reconnect failed: %v", reconnErr)
 	}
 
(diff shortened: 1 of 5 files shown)


sunapi386:
Follow-up commit `5d0a328`: threads `apiBaseURL` through `reclaimStaleNodeByHostname()`.

The original commit hardcoded `authServiceURL` in the reclaim function, so on self-hosted backends (where `deviceConfig.APIBaseURL` differs from the default `https://aceteam.ai`) the reclaim would silently target the wrong server and the churn fix would be a no-op. All 5 call sites now pass their resolved `apiBaseURL` through.

**Residual note on Attempt 1 (IP-preserving reconnect):** `ReconnectWithAuthKey` reuses the existing machine key with a fresh authkey. Whether Headscale reuses the same node ID or mints a new one depends on Headscale's key-matching behavior. If it mints a new ID, there is still residual churn on the "success" path. This PR is still a net improvement -- it makes the ClearState fallback rarer (3 retries) and cleans up stale nodes when it does happen -- but full elimination of churn may require verifying Headscale's authkey-vs-machine-key identity model.

PR #272, opened 2026-06-21
Co-browse: managed Chromium over CDP with human handoff (#4079)

sunapi386:
Node side of human-in-the-loop co-browsing (#4079; cockpit render seam #3967).

## What
A managed headed-Chromium session on a node, driven by the AI agent over CDP, with an explicit human-in-the-loop handoff so the human can log in / do 2FA / take over.

## How
- `internal/platform/cobrowse.go` — `CobrowseManager` singleton (mirrors `GetVNCManager`) owning one long-lived headed Chromium, launched with a persistent `--user-data-dir` profile and `--remote-debugging-port` (CloakBrowser-style, like td_serve.py). Driven via **raw CDP over gorilla/websocket** (already a dep) — `Page.navigate` + `Page.captureScreenshot` only. No chromedp / heavy dependency added.
- **Handoff is enforced, not advisory:** driver state machine (`ai`/`human`). While `human`, AI `Navigate`/`Screenshot` return `ErrHandedOff`. `Resume` returns control. State is queryable via `Status`.
- `internal/jobs/cobrowse.go` — `COBROWSE` job handler dispatching start/navigate/handoff/resume/status/screenshot to the manager.
- Registered `JobTypeCobrowse` in `worker/job.go`, `handler_adapter.go` (the real `runWork` path) and `cmd/job_handlers.go` (test path).

## Live view
The browser is **headed on the node display**, so the existing VNC WebSocket bridge (`/vnc/...` → websockify, see `cmd/serve.go`/`cmd/work.go`) already streams it to the cockpit. No new streaming transport — chosen over a CDP screencast because the desktop stream already exists end-to-end and also shows the human's takeover.

## Tests
`go build ./...`, `go vet ./...` pass. `internal/jobs/cobrowse_test.go` covers the state-machine contract (missing/unknown action, ErrNotStarted for actions without a session, status answerable when not running, distinct error sentinels). `internal/jobs`, `internal/worker`, `internal/platform` test packages pass.

Pairs with the aceteam `cobrowse_*` MCP tools PR.

### DIFF
--- a/cmd/job_handlers.go
+++ b/cmd/job_handlers.go
@@ -69,5 +69,6 @@ func init() {
 		"IOS_BUILD":          jobs.NewIOSBuildHandler(""),
 		"ANDROID_BUILD":      jobs.NewAndroidBuildHandler(""),
 		"GOMOBILE_BUILD":     jobs.NewGomobileBuildHandler(""),
+		"COBROWSE":           jobs.NewCobrowseHandler(),
 	}
 }
--- a/cmd/work.go
+++ b/cmd/work.go
@@ -167,6 +167,12 @@ func runWork(cmd *cobra.Command, args []string) {
 	go func() {
 		<-sigs
 		fmt.Println("\n   - Received shutdown signal...")
+		// Kill any managed co-browse Chromium so a headed browser does not orphan
+		// across worker restarts (the persistent profile keeps logins; only the
+		// live process is torn down). No-op when no session was ever started.
+		if err := platform.GetCobrowseManager().Stop(); err != nil {
+			fmt.Fprintf(os.Stderr, "   - Warning: co-browse browser stop: %v\n", err)
+		}
 		cancel()
 		go func() {
 			select {
--- a/internal/jobs/cobrowse.go
+++ b/internal/jobs/cobrowse.go
@@ -1,0 +1,111 @@
+// internal/jobs/cobrowse.go
+//
+// Co-browse job handler (issue #4079). Bridges the node's Redis job stream to
+// the persistent CobrowseManager so the AI agent can drive a managed headed
+// Chromium over CDP and hand control to a human for login / 2FA.
+//
+// Each job is one short action against the long-lived browser owned by
+// platform.GetCobrowseManager(); the browser itself persists between jobs.
+// The handler returns a JSON document as its output bytes so the backend MCP
+// tool can parse a structured result (mirroring the VNC handlers' wire shape).
+package jobs
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/aceteam-ai/citadel-cli/internal/nexus"
+	"github.com/aceteam-ai/citadel-cli/internal/platform"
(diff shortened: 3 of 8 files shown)


sunapi386:
Review + hardening pass (driving to ready)

## Reviewed
- Full diff vs main, PR #272, aceteam-ai/aceteam#4079, and the paired `cobrowse_*` MCP tool schemas.
- Confirmed the cross-repo wire contract matches: MCP params (`node_id` + optional `url` for start, `url` for navigate, bare for the rest) map cleanly onto the handlers payload (`action`/`url`/`profile`), the `ai`/`human` driver states, the `running`/`url` status shape, and the screenshot-refused-during-handoff guardrail. No field/action/shape mismatch. (`profile` is not exposed by the MCP tool yet; the handler treats it as optional, so that is forward-compatible, not a break.)

## Fixed
- **Process-death / zombie bug (load-bearing):** `isRunningLocked` only checked `cmd.Process != nil`, which stays non-nil for the life of the struct, so a crashed Chromium was reported as running forever and was never reaped. Added a reaper goroutine that owns `cmd.Wait()` and closes an `exited` channel; `isRunningLocked` now consults it. The `Start` reuse path therefore correctly relaunches after a crash instead of navigating a dead browser.
- **`Stop` lifecycle:** no longer double-`Wait()`s (the reaper owns `Wait`). It kills, waits bounded (5s) on the reaper to confirm reaping, and is safe to call when not running and twice.
- **Killed on shutdown:** wired `CobrowseManager.Stop()` into the worker SIGINT/SIGTERM handler in `cmd/work.go` so a headed Chromium does not orphan across worker restarts (the persistent profile keeps logins; only the live process is torn down).
- Documented the intentional per-node action serialization on the manager mutex.

## Tests / verification
- New scoped `internal/platform/cobrowse_test.go`: reaper exit-detection (POSIX `sleep` stand-in, no Chromium/display needed), `Stop` kills a live process + resets driver, double-`Stop` safe.
- `go build ./...` clean; `go vet ./internal/platform ./internal/jobs ./internal/worker ./cmd` clean.
- Scoped tests green: `internal/platform` (cobrowse), `internal/jobs` (cobrowse), `internal/worker`. No `go test ./...` and no tmux/terminal test packages were run.

## Needs a live Chromium + display to verify (cannot be exercised in-agent)
1. `DISPLAY=:0` actually landing the headed browser on the x11vnc-streamed display (VNC visibility in the cockpit).
2. End-to-end CDP round-trip: `Page.navigate` then `Page.captureScreenshot` returning valid base64 against a real DevTools endpoint.
3. Handoff visibility: during handoff the agent is blind (screenshot refused by design) and polls `cobrowse_status` (URL only) to know when to `resume` -- confirm `status` is the intended resume trigger for the human-watching flow.
4. Timing nuance: `Page.navigate` returns on navigation initiation, not load-complete, so a screenshot issued in the same instant could catch a pre-load page; across separate MCP calls this is a non-issue in practice.

PR #298, opened 2026-06-23
Closes #293: chat connection status + broker-less discovery

### LINKED ISSUE #293: Chat (Alt+3): stuck 'connecting…', needs a connection settings/status page + broker-less peer discovery
### 1. 'connecting…' never resolves
The Alt+3 Chat pane sits on 'connecting…'. Investigate the chat WS/connection lifecycle in `internal/tui/controlcenter/chat_page.go` — why it never reaches connected (handshake? endpoint? auth? the same relay/VPN path issues?).

### 2. A chat connection settings/status page
Add a settings/status view showing **which WSS endpoint we're connected to** and connection health. **Do NOT expose that the transport is Redis** — surface the user-facing WSS/connection, not the internal broker.

### 3. Broker-less peer discovery (design note from Jason)
We don't currently have a way to *discover* chat peers without a central broker. Insight: use **Headscale / nexus** to enumerate the IPs visible to our org, and **probe whether each speaks the chat protocol** — i.e. discover peers over the mesh instead of via a broker. Scope this as the discovery mechanism.

## Related
- aceteam-ai/aceteam#3492, the terminal/relay work.

sunapi386:
## Context

The Control Center **Chat** pane (Alt+3) was stuck on `connecting…` indefinitely. Issue #293 asks to (1) diagnose the stuck state, (2) surface which **WSS** endpoint the chat uses + its health **without** revealing that the transport is Redis, and (3) design broker-less peer discovery over the Headscale/nexus mesh.

## Connecting diagnosis & fix

**Root cause (in-scope, fixed):** `ChatPage.connect()` set the message view to `[green]Connected` *before* the blocking `chat.Client.Connect()` call and then discarded that call's error with `_ = client.Connect(ctx)`. The peers sidebar was seeded with a static `connecting…` placeholder that was only ever cleared by a presence callback. So whenever the real-time handshake failed (endpoint unreachable, auth rejected, or subscribe error), no presence ever arrived: the sidebar stayed on `connecting…` forever while the main view falsely showed `Connected`. A silent, permanent half-connected state.

**External unknown (why this is a DRAFT):** whether the WSS endpoint (`/api/fabric/redis/ws` on the API host) is actually deployed and reachable in a given environment can't be verified from here — that requires the running TUI + backend. The fix makes the UI tell the truth about whatever the handshake actually does, rather than masking it.

### Changes

- `chat.Client` gains a real connected signal: `OnConnect(fn)` fires **once, after** the WebSocket handshake **and** channel subscribe succeed; `IsConnected()` is a read-only transport passthrough. *(Touches `internal/chat/client.go` — a third file beyond the two TUI files in scope. This is non-TUI, so there's no collision with parallel TUI agents; calling it out explicitly. A pure time-based heuristic was rejected because the 10s handshake timeout makes a timer report "connected" mid-dial.)*
- `chat_page.go` now drives a `connecting → connected | error` state machine: stays in **connecting** until `OnConnect`; on error from `Connect()` it shows the **real error message** and a red `offline` sidebar instead of an eternal `connecting…`.

## WSS status UI (Redis not exposed)

A one-line status bar between the message view and input shows the endpoint + health:

```
● connecting…  wss://aceteam.ai
● connected    wss://aceteam.ai
● error        wss://aceteam.ai   <real error>
```

The label comes from `chat.SanitizeEndpoint`, which returns `scheme://host` only (https→wss, http→ws) and **drops the path**. The internal Redis pub/sub transport path is never surfaced. A unit test asserts the sanitized output never contains `redis` or a path.

## Broker-less discovery (design + stub)

`internal/chat/discovery.go` + `docs/chat-brokerless-discovery.md`. Chat is broker-mediated today (central WSS fan-out). Broker-less discovery removes the central directory: every org node is already on the same Headscale mesh, so a node can enumerate org peers locally and probe them.

| Step               | Status      | Notes |
| ------------------ | ----------- | ----- |
| Enumerate          | Implemented | `network.GetGlobalPeers` — org-scoped via Headscale user `org_<id>`. |
| Reachability probe | Implemented | `network.PingPeer` over the mesh; proves routable, **not** chat-able. |
| Protocol probe     | Stub        | No per-node chat listener exists yet to probe against; `ProbeResult.SpeaksChat` is always false (asserted by test). |

`DiscoverPeers(ctx)` ships the enumerate+probe plumbing; the doc spells out the follow-up (per-node listener, real protocol handshake, direct mesh routing with broker fallback).

## Test plan

| Area | Coverage |
| ---- | -------- |
| `SanitizeEndpoint` | https/http/wss conversion, path stripping, port retention, empty/invalid input |
| No-transport-leak guard | sanitized output never contains `redis` or a path |
| `EndpointURL` passthrough | client returns sanitized host |
| Discovery invariant | `SpeaksChat` defaults false (no listener yet) |

`go build ./...`, `go vet ./...`, `go test ./...` all clean. **Live verification of the connecting fix needs the running TUI + backend** and is not done here.

## Scope

Edits confined to `internal/tui/controlcenter/chat_page.go` and the `internal/chat` helpers (`client.go`, `message.go`, new `discovery.go`/`discovery_test.go`) plus a design doc. No pane registration, other panes, or settings pane touched.

## Related

Closes #293.

### DIFF
--- a/docs/chat-brokerless-discovery.md
+++ b/docs/chat-brokerless-discovery.md
@@ -1,0 +1,105 @@
+# Chat: connection status & broker-less peer discovery
+
+This document covers the Control Center **Chat** pane (Alt+3): how it connects,
+the connection-status indicator, and the design for broker-less peer discovery
+over the Headscale/nexus mesh.
+
+## Background: the "connecting…" bug (issue #293)
+
+The Chat pane connects lazily on first activation. The original lifecycle
+optimistically set the message view to `Connected` **before** the blocking
+`chat.Client.Connect()` call, and then discarded that call's error
+(`_ = client.Connect(ctx)`). The peers sidebar was seeded with a static
+`connecting…` placeholder that was only ever cleared by a presence callback.
+
+The result: if the real-time handshake failed (endpoint unreachable, auth
+rejected, subscribe error) no presence ever arrived, so the sidebar stayed on
+`connecting…` forever while the main view falsely claimed `Connected`. The user
+saw a permanent, silent half-connected state.
+
+### Fix
+
+The chat client now exposes a real connected signal:
+
+- `chat.Client.OnConnect(fn)` — fired exactly once, **after** the WebSocket
+  handshake and channel subscribe both succeed (`internal/chat/client.go`).
+- `chat.Client.IsConnected()` — read-only passthrough to the transport.
+
+The Chat pane (`internal/tui/controlcenter/chat_page.go`) drives a three-state
+machine — `connecting → connected | error`:
+
+- It stays in **connecting** until `OnConnect` fires.
+- On `OnConnect` it flips to **connected**, clears the sidebar placeholder, and
+  marks the page connected.
+- If `Connect()` returns an error before a successful connect, it flips to
+  **error**, shows the real error message in the message view and a red
+  `offline` sidebar, instead of an eternal `connecting…`.
+
+## Connection status indicator
+
+A one-line status bar sits between the message view and the input box. It shows
+the **WSS endpoint** the chat is using and its health:
+
+```
+● connecting…  wss://aceteam.ai
+● connected    wss://aceteam.ai
+● error        wss://aceteam.ai   <real error>
+```
+
+The endpoint label is produced by `chat.SanitizeEndpoint`, which converts the
+API base URL to `scheme://host` (https→wss, http→ws) and **deliberately drops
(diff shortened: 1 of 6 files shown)


sunapi386:
Superseded by #301, which integrates #293/#294/#295 as one coherent branch (de-dupes the overlapping `IsConnected`/`telemetry.go` edits and resolves the shared `controlcenter.go` wiring). Closing in favor of that. The issue this PR closed will be closed by #301.

Issue #295, opened 2026-06-23
citadel-cli: Settings page — manage anon-data opt-out + show connection/WSS status


sunapi386:
## Goal
A **Settings** view in the citadel-cli TUI (a new pane / Alt-key) where the node owner can manage preferences and see connection status.

## Contents
1. **Anonymous data collection** — toggle to opt in/out of the remote-debug activity streaming (#294). Clear copy on what's collected (anonymous debug events) and that it's used to debug remotely.
2. **Connection status** — show **which WSS endpoint** the chat/control connection is on + health (from #293). **Do NOT expose that the transport is Redis** — surface the user-facing WSS/connection only.
3. (room to grow: VPN/mesh status, terminal/VNC bind status — the #288 loud bind log could surface here.)

## Related
- #294 (the telemetry it toggles), #293 (chat connection status — could be folded in or linked).

Issue #291, opened 2026-06-23
TUI crashes: panic in tailscale logtail/filch (slice bounds out of range) — disable the log uploader


sunapi386:
## Crash (observed in the Alt+4 Proxmox view, but it's in a background goroutine)

```
panic: runtime error: slice bounds out of range [51:0]
tailscale.com/logtail/filch.(*Filch).unreadReadBuffer(...) filch.go:132
tailscale.com/logtail/filch.(*Filch).TryReadLine(...)       filch.go:181
tailscale.com/logtail.(*Logger).drainPending(...)           logtail.go:309
tailscale.com/logtail.(*Logger).uploading(...)              logtail.go:372
created by tailscale.com/logtail.NewLogger
```
The whole TUI app dies ('it seemed to have died by itself').

## Root cause
A bug in **tailscale v1.94.0's logtail/filch** (the on-disk log buffer the telemetry uploader reads). We set tsnet's `Logf: func(...) {}` (server.go:115) — but that only suppresses verbose logs; the **logtail uploader** (which ships node logs to log.tailscale.io) still runs the buggy filch path.

## Fix (prefer the least-invasive, verified one)
citadel is an embedded node — it does NOT need tailscale's telemetry log upload. Disable the logtail uploader / filch persistence:
- Investigate the tsnet/tailscale knob to disable logtail (e.g. `tsnet.Server` log options, `TS_NO_LOGS_NO_SUPPORT`, or a no-op logtail). 
- If no clean disable exists, consider bumping tailscale past v1.94.0 (only if a release fixes the filch bounds bug) or a `patches/` patch (the repo already patches tcell).
Pick the safest fix that keeps VPN connectivity intact; build/vet/test; note full verification needs running on a node.

## Related
- aceteam-ai/aceteam #3492 (mobile/terminal), surfaced 2026-06-22.

PR #300, opened 2026-06-23
feat: stream activity feed to events log for remote debugging (Closes #294)

### LINKED ISSUE #294: Stream citadel-cli activity to the Supabase events log for remote debugging (anonymous, opt-out)
## Goal
Stream the citadel-cli **activity page** (node activity / debug events) to the **Supabase events log** so the team can debug nodes **remotely** without asking the user to copy logs. Treat it as **anonymous debug telemetry**.

## Requirements
- Emit activity/debug events to a Supabase events-log table (reuse an existing events/usage table if one fits, e.g. the heartbeat/usage path, rather than a new table — coordinate with the backend; DB is owner-managed).
- **Anonymous** — strip/avoid PII; it's for debugging node behavior, not user identification.
- **Opt-out** — the node owner can disable anon data collection entirely (default-on for debuggability, or default per policy — decide with Jason). When opted out, nothing is streamed.
- Low overhead / fire-and-forget; never block or crash the TUI if the stream fails (cf. the logtail crash #291).

## Manage it via a settings page
See #295 — the opt-out toggle lives there.

## Related
- #295 (settings page), #291 (don't repeat the crash-on-telemetry pattern), aceteam-ai/aceteam events/usage infra.

sunapi386:
## Context

Closes #294. Related #295.

Field nodes we can't SSH into are hard to debug — the only window into what a node is doing is its on-screen **Activity** feed in the control-center TUI. This PR streams that feed to the AceTeam control plane so the activity log is visible remotely (the "Supabase events log"), turning every node into something we can debug after the fact.

The change is **anonymous** (node/debug context only, no user PII) and **opt-out** (a persisted flag the operator can flip off). Emission is **fire-and-forget and crash-safe** so it can never block or panic the TUI — directly informed by the logtail crash (#291).

## Activity source

The TUI activity feed is fed by a single chokepoint: `ControlCenter.AddActivity(level, message)` in `internal/tui/controlcenter/controlcenter.go`. Every status line (VPN connect/reconnect, worker start, job output, peer reachability, etc.) flows through it. Hooking emission there captures the entire feed without touching any TUI pane.

## Emission path — reused backend channel (no direct Supabase)

`AddActivity` now calls `telemetry.Emit(level, message)`. The new `internal/telemetry` package writes each entry to a Redis stream **via the same device-token-authenticated Redis API proxy the heartbeat already uses** (`redisapi.Client.StreamAdd` → `POST /api/fabric/redis/streams/add`). Citadel never touches Supabase directly.

- Emitter is a configured singleton, wired in `ccStartWorker` right beside the heartbeat publisher, reusing `apiSource.Client()` and the same node/org context (nodeName, headscaleNodeID, orgID, version).
- Before `Configure` runs (early pre-worker activity) `Emit` is a silent no-op.
- Stream: **`node:activity:stream`**, fields mirroring the heartbeat shape, `MAXLEN ~ 10000`.
- Uses durable `StreamAdd` (not ephemeral pub/sub) so a backend consumer can drain it reliably.

## Opt-out flag: `anon_telemetry_enabled`

`internal/config/telemetry.go` adds the flag, mirroring `permissions.go` exactly (YAML file in the config dir, opt-out/default-true semantics, partial-file-preserves-defaults).

- File: `telemetry.yaml`, key: **`anon_telemetry_enabled`**, **default `true`**.
- Default-on rationale (commented in code): the feed is anonymous and remote visibility is the primary way to debug nodes we can't reach. User-facing disclosure of the default-on behavior is #295 / `citadel init`'s job.
- `LoadTelemetry` / `SaveTelemetry` give #295's settings pane a symmetric API to toggle the flag.
- The flag is **re-read per emit** so a runtime toggle takes effect without restart (same pattern as permissions being reloaded each heartbeat). The read is off the TUI thread. _(If activity volume warrants, this could be cached with a short TTL; left simple for now.)_

## Guarantees

**Anonymity.** The payload carries only node/debug context: `nodeId`, `headscaleNodeId`, `orgId`, `version`, `timestamp`, `level`, `message`. User-identity fields (email/name) are **structurally excluded** from the payload. A call-site audit of `AddActivity` found messages are operational status lines that do not interpolate user identity; messages are free-form text, so we claim the structured fields are PII-free and the messages are operational, not that arbitrary message text is provably PII-free.

**Crash-safety / fire-and-forget.** `Emit` spawns a guarded goroutine with `recover()`; it never blocks the caller and never panics into the TUI. A single emission is bounded by a 5s context timeout. Failures are silently dropped — telemetry is best-effort, no retries/batching.

## Test plan

| Area | Coverage |
| --- | --- |
| Gating | flag off → zero emissions; flag on → exactly one; default (no file) → enabled → emits |
| Anonymization | payload contains required node/debug fields and **no** PII-like keys (email/username/token/secret/password) |
| Crash-safety | `Emit` unconfigured → no panic; `Configure(nil)` clears the emitter |
| Config flag | default-on, save/load round-trip, disabled-from-file, invalid-YAML→defaults, nested-dir create |

`go build ./...`, `go vet ./...`, and `go test ./...` are all clean. Gating/anonymization are tested against `emit()` synchronously (injected fake sink, temp config dir) to avoid goroutine flakiness.

## Backend / DB follow-up required (why this PR is a DRAFT)

The citadel side is complete, but nothing reaches the events log until the backend can accept and persist this stream. Required, owner-managed work:

1. **Allowlist the new stream name.** `app/api/fabric/redis/streams/add/route.ts` has a strict allowlist — `ALLOWED_STREAM_PATTERNS` currently only permits `^node:status:stream$`. Until `node:activity:stream` is added, every emit returns 400 and is silently dropped. (The required `device_redis:streams` scope is already on the device token via heartbeat, so no scope change is needed.)
2. **Add a Python consumer** to drain `node:activity:stream` (mirroring `worker/node_status/worker.py`, which drains `node:status:stream`).
3. **Create the events table** (DB is owner-managed — **not created here**). Suggested `node_activity_events` columns: `id` (uuid pk), `node_id` (text), `headscale_node_id` (text), `organization_id` (uuid fk → organizations, `ON UPDATE CASCADE`), `version` (text), `level` (text/enum), `message` (text), `reported_at` (timestamptz), `created_at` (timestamptz default now()). Existing `agent_activity_events` is agent-scoped (FK to agents) and not suitable; `fabric_node_status_history` is heartbeat-specific.

### DIFF
--- a/cmd/controlcenter.go
+++ b/cmd/controlcenter.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aceteam-ai/citadel-cli/internal/platform"
 	pmx "github.com/aceteam-ai/citadel-cli/internal/proxmox"
 	"github.com/aceteam-ai/citadel-cli/internal/status"
+	"github.com/aceteam-ai/citadel-cli/internal/telemetry"
 	"github.com/aceteam-ai/citadel-cli/internal/terminal"
 	"github.com/aceteam-ai/citadel-cli/internal/tui"
 	"github.com/aceteam-ai/citadel-cli/internal/tui/controlcenter"
@@ -1167,6 +1168,19 @@ func runTUIWorker(ctx context.Context, activityFn func(level, msg string)) error
 
 		if orgID != "" {
 			if apiSource, ok := source.(*worker.APISource); ok {
+				// Wire anonymous activity telemetry through the same
+				// authenticated Redis API client the heartbeat uses. Gated
+				// at emit time by the anon_telemetry_enabled flag; payloads
+				// carry only node/debug context, never user PII.
+				telemetry.Configure(
+					apiSource.Client(),
+					platform.ConfigDir(),
+					nodeName,
+					headscaleNodeID,
+					orgID,
+					Version,
+				)
+
 				apiPublisher, err := heartbeat.NewAPIPublisher(heartbeat.APIPublisherConfig{
 					Client:          apiSource.Client(),
 					NodeID:          nodeName,
--- a/internal/config/telemetry.go
+++ b/internal/config/telemetry.go
@@ -1,0 +1,68 @@
+package config
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Telemetry controls anonymous activity-event streaming used for remote
+// debugging. It follows the same opt-out (default-true) model as Permissions:
+// the node streams its operational activity feed to the AceTeam control plane
+// so operators and the platform can debug node issues remotely, and the
+// operator may turn it off.
+type Telemetry struct {
+	// AnonTelemetryEnabled gates ALL activity-event emission. When false, no
+	// activity events leave the node. Defaults to true (opt-out) so that nodes
+	// are debuggable out of the box; the settings pane (#295) and `citadel
+	// init` are responsible for disclosing this default-on behavior to the
+	// operator and offering the toggle.
+	AnonTelemetryEnabled bool `yaml:"anon_telemetry_enabled" json:"anon_telemetry_enabled"`
+}
+
+const telemetryFile = "telemetry.yaml"
+
+// DefaultTelemetry returns a Telemetry struct with anonymous telemetry enabled.
+// Default-on is intentional: the activity feed is anonymous (node/debug context
+// only, no user PII) and remote visibility into the activity log is the primary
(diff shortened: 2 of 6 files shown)


sunapi386:
Superseded by #301, which integrates #293/#294/#295 as one coherent branch (de-dupes the overlapping `IsConnected`/`telemetry.go` edits and resolves the shared `controlcenter.go` wiring). Closing in favor of that. The issue this PR closed will be closed by #301.

PR #342, opened 2026-06-24
feat(module): install service modules from any standardized git repo

sunapi386:
## Context

Today a Citadel service module can only be installed from the central `aceteam-ai/citadel-services` catalog. This PR adds a generic **module source** mechanism so a module can be installed from **any standardized git repo**, via both the CLI and the TUI control center.

A module repo self-describes via a top-level `citadel/` directory containing `service.yaml` + `compose.yml`, with the compose `image:` pointing at that repo's own container registry. The CLI never hardcodes or rewrites a module's image — the module repo owns it.

## The mechanism

A "source" is one of:

| Form | Example | Resolves to |
|------|---------|-------------|
| catalog name | `vllm` | central catalog (existing path) |
| `owner/repo` | `owner/repo` | `https://github.com/owner/repo.git` |
| `owner/repo@ref` | `owner/repo@v1.2.0`, `owner/repo@<sha>` | clone + checkout tag/branch/**SHA** |
| full git URL | `https://git.example/m.git`, `git@host:owner/repo.git` | cloned directly |

External sources are shallow-cloned into a per-source cache under `ConfigDir()/modules/<owner-repo[@ref]>/` (mirroring `catalog.Update()`'s clone/pull style). The manifest is located at `citadel/service.yaml` (preferred) or `service.yaml` (fallback); compose at `citadel/compose.yml` then `compose.yml`.

### Private-repo credentials caveat

Cloning a **private** source repo requires the node to have git credentials. The clone-failure error now names the host (e.g. `github.com`) and the concrete fixes: a `GITHUB_TOKEN` env var, an SSH key for that host, or a git credential helper (`git credential-store` / `gh auth login`). Public repos work for everyone; private repos work only for nodes with credentials.

## New command surface

```
citadel module install <source> [--set KEY=VALUE ...] [--yes]
citadel module list
citadel module info <source>
```

`citadel module` is a new **top-level** group — deliberately NOT under `citadel service install` (the systemd service-install command) nor `citadel service catalog`. Catalog-name sources delegate to the existing catalog install flow (so host-provisioned / wechat guidance does not diverge).

## Routing tags — third-party engines become routable with no CLI change

`service.yaml` gains a `node_tags` field: namespaced `key:value` routing tags (e.g. `engine:tei`, `task:embedding`, `model:gte-multilingual-base`). On install these are merged into the node manifest's `node.tags` (deduped), which is how Citadel's capability-based queue routing picks them up. This means a third-party engine module (e.g. a TEI embedding service) declares its own routing tags and becomes routable **without** editing the CLI's hardcoded tag map. `node_tags` is distinct from the existing display/search `tags`; the hardcoded `serviceTags` map is kept intact for embedded/catalog services.

## Provenance + lockfile (reproducible, tamper-evident nodes)

After a successful resolve+install, a per-module entry is recorded into `ConfigDir()/modules.lock` (YAML): `name`, normalized `source`, requested `ref`, resolved git `commit` (`git rev-parse HEAD`), and the image reference(s) parsed from the compose, each with a **best-effort** `sha256:` digest (via `docker image inspect` / `docker manifest inspect`, time-bounded to ~3s, omitted on any failure — never fails the install). `citadel module list` merges this with the node manifest to show `MODULE  SOURCE@commit  IMAGE[@digest]`; services without a lockfile entry (catalog/embedded) still appear. The non-`--yes` pre-install prompt now prints the resolved **commit + image(s)** so it is a real TOFU/provenance confirmation.

## Conflict detection on install

In addition to the existing host-port conflict check, `InstallFromManifest` now parses `container_name:` from the resolved compose and refuses the install if a container with that name already exists and is not this module (a same-module reinstall is already blocked upstream by the manifest `hasService` check), pointing at the `docker rm -f` / compose-override escape hatch. Best-effort: skipped silently if docker is unavailable. Note: this now also applies to catalog installs (consistent, low-risk behavior change).

## Schema versioning

`service.yaml` gains `schema_version` (int). A value newer than the CLI's `CurrentSchemaVersion` (1) prints a forward-compat **warning** (some fields may be ignored) but never fails — older CLIs degrade gracefully against future modules. The warning is computed by a pure `SchemaWarning()` and printed from the cmd layer (the catalog package stays import-clean).

## `requires` preflight

GPU / VRAM / arch requirements declared in the manifest are enforced for module installs via the shared `InstallFromManifest` checks (arch mismatch and missing-GPU are hard errors) — the same path catalog installs use. No new code needed here.

## TUI

A new "Install Module" page in the control center (wired only when callbacks are injected from cmd, like the WhatsApp page). Two-phase `tview.Form`: type a source → Resolve (clones off the UI goroutine) → fill required config → Install. All config is collected up front and passed as overrides, so the **non-interactive** installer path is used — the TUI never reads `os.Stdin`. The TUI path has full parity with the CLI: already-installed guard, `node_tags` merge, and lockfile provenance recording.

## Implementation notes

- `Install()` refactored into `InstallFromManifest(manifest, composeSrcPath, servicesDir, overrides, interactive)`. The old `Install(name, ...)` signature is unchanged (thin wrapper → loads from cache → `interactive=true`). `resolveConfig` only prompts on stdin when `interactive==true`; otherwise a required var with no override/default is a returned error. Empty `composeSrcPath` → `ErrNotInstallable` (preserves the host-provisioned wechat case).
- `ResolveSource` returns a `ResolvedModule{Manifest, ComposePath, CacheDir, Commit, Images}` struct.
- Pure decision logic (`looksLikeSHA`, `pickCloneStrategy`, `parseComposeImages`, `parseComposeContainerName`, `cloneErrorHost`, `SchemaWarning`) is split out and table-tested with no network.

## Files

- `internal/catalog/source.go` — `ParseSource`, `ResolveSource`, clone/update (incl. SHA fetch+checkout), pure decision helpers, schema warning
- `internal/catalog/lockfile.go` (new) — `modules.lock` load/upsert, best-effort image digest, container-name conflict check
- `internal/catalog/source_test.go` — table-driven tests (ParseSource, SHA/strategy detection, compose parsing, host extraction, lockfile upsert, schema warning) — no network
- `internal/catalog/catalog.go` — `NodeTags` + `SchemaVersion` fields on `ServiceManifest`, `CurrentSchemaVersion`
- `internal/catalog/install.go` — `InstallFromManifest`, `interactive` flag, container-name conflict check
- `cmd/module.go` — `citadel module` group, provenance recording, `buildModuleInstallCallbacks()`
- `cmd/manifest.go` — `addServiceToManifestWithTags` (wrapper keeps `addServiceToManifest`)
- `internal/tui/controlcenter/module_page.go` (new) — Install Module page + callback types
- `internal/tui/controlcenter/controlcenter.go`, `cmd/controlcenter.go` — wire the page

## Purely additive

No changes to `cmd/whatsapp.go`, `internal/whatsapp/*`, `services/embed.go`, `services/compose/*`, or any `ServiceMap` entry.

## Testing

```
go build ./...        # clean
go vet ./...          # clean
go test ./internal/catalog/...   # PASS
```

Per repo policy, the full `go test ./...` was not run from this environment (tmux/pty suites). Build + vet cover compilation of all packages.

### Manual TUI test (why this is a draft)

1. `citadel dashboard` (control center)
2. Tab to **Install Module**
3. Type a source (e.g. `owner/repo` for a public standardized repo) → **Resolve**
4. Fill any required config fields (`*` = required) → **Install**
5. Confirm success line and that `citadel module list` shows `source@commit` + image

## Trust & safety (install = root-equivalent access)

Installing a module runs `docker compose up` of arbitrary compose, which is root-equivalent (Docker-level) access to the node. Before merge, the install flow gained a trust/safety layer:

**Untrusted-source warning.** External sources are checked against a local allowlist (`citadel module trust <pattern>` → `ConfigDir()/trusted_sources.yaml`). A non-trusted source prints a prominent warning that the install runs an arbitrary container with host-root-equivalent access. Patterns: exact `owner/repo`, `owner/*`, or a bare host (`github.com`). Catalog sources are **Tier 0** (first-party, always trusted, never scanned/gated). Trust suppresses the warning **only** — it does not relax the privilege gate.

**Compose risk-scan** (`internal/catalog/risk.go`, pure + table-tested). The resolved compose is scanned for dangerous directives:
- **Critical:** `privileged: true`; a `/var/run/docker.sock` (or `/run/docker.sock`) bind mount; `cap_add` containing `ALL` or `SYS_ADMIN`.
- **High:** `network_mode: host`, `pid: host`, `ipc: host`; host bind-mounts of sensitive paths (`/`, `/etc`, `/root`, `/var/run`, `/home`, `~`, …, boundary-aware so `/etcetera` ≠ `/etc`, named volumes excluded); any `devices:` host-device passthrough.

Findings are shown at install time (Critical red, High yellow) in both the CLI and the TUI form.

**Privilege gate — `--yes` vs `--allow-privileged`.** The gate lives in the shared core (`InstallFromManifest`), so CLI and TUI are protected identically:

| Critical risk? | `--allow-privileged` | `--yes` | Result |
|---|---|---|---|
| yes | no | any | **REFUSED** (names the critical directives + how to override), even under `--yes` |
| yes | yes | no | warning + findings, ordinary `[y/N]` prompt |
| yes | yes | yes | proceeds |
| no | — | no | warning (if untrusted) + High findings, ordinary `[y/N]` prompt |
| no | — | yes | proceeds |

`--yes` bypasses only the ordinary confirmation prompt; it does **not** bypass a Critical finding. Trusted + non-critical + no `--yes` still prompts (trust silences the scary banner, not the confirmation). The TUI equivalent of `--allow-privileged` is an explicit "Allow privileged (root-equivalent) access" checkbox that only appears when a Critical risk is present; the non-interactive TUI install path hard-fails on Critical-without-opt-in (no silent privileged install, no stdin). The core gate fails **closed**: if the compose can't be read for the scan, the install is refused rather than proceeding. Catalog (Tier-0) installs pass `allowPrivileged=true` and are exempt (they have no override flag) — on both CLI and TUI.

**`citadel module trust` surface:**
```
citadel module trust <pattern>      # add owner/repo | owner/* | host to the allowlist
citadel module untrust <pattern>    # remove a pattern
citadel module trust --list         # list patterns
citadel module trusted              # list patterns
```

**Git-ref injection guard.** A source ref is passed positionally to `git clone --branch` / `git fetch origin <ref>`, so a value beginning with `-` (or containing whitespace/control chars) is rejected (`validateRef`) to prevent git-argument injection (e.g. `--upload-pack=…`).

New files: `internal/catalog/risk.go` (+ `risk_test.go`), `internal/catalog/trust.go` (+ `trust_test.go`), `cmd/module_trust.go`. `InstallFromManifest` gained an `allowPrivileged` parameter (legacy `Install` wrapper passes `true`).

## Deferred (explicit follow-ups)

- Cosign/sigstore signature verification + verified-publisher allowlist.
- Semver / channel resolution (`@^1.2`, `@stable`).
- `module update` / `module list --outdated` / cache GC / auto-rollback on failed `health_check`.
- Central index repo + `citadel module search` discovery.

*Generated by Claude*

### DIFF
--- a/cmd/controlcenter.go
+++ b/cmd/controlcenter.go
@@ -240,7 +240,7 @@ func runControlCenter() {
 		// longer shows a permanent "device authorization required" status.
 		Chat:               buildChatConfig(),
 		ChatConfigProvider: buildChatConfig,
-		Proxmox:   buildProxmoxConfig(),
+		Proxmox:            buildProxmoxConfig(),
 		Settings: controlcenter.SettingsCallbacks{
 			LoadTelemetry: func() *config.Telemetry {
 				return config.LoadTelemetry(platform.ConfigDir())
@@ -255,7 +255,8 @@ func runControlCenter() {
 				return config.SaveKeepAwake(platform.ConfigDir(), k)
 			},
 		},
-		WhatsApp: buildWhatsAppCallbacks(),
+		WhatsApp:      buildWhatsAppCallbacks(),
+		ModuleInstall: buildModuleInstallCallbacks(),
 	}
 
 	cc := controlcenter.New(cfg)
--- a/cmd/manifest.go
+++ b/cmd/manifest.go
@@ -14,7 +14,7 @@ import (
 // Service defines a single managed service.
 type Service struct {
 	Name        string `yaml:"name"`
-	Type        string `yaml:"type,omitempty"`        // "native" or "docker" (default: auto-detect)
+	Type        string `yaml:"type,omitempty"`         // "native" or "docker" (default: auto-detect)
 	ComposeFile string `yaml:"compose_file,omitempty"` // For docker services
 	Port        int    `yaml:"port,omitempty"`         // For native services
 }
@@ -28,7 +28,7 @@ type ManifestCapabilities struct {
 
 // ManifestGPU describes a GPU declared in the manifest.
 type ManifestGPU struct {
-	Name   string `yaml:"name"`             // e.g. "NVIDIA GeForce RTX 3090"
+	Name   string `yaml:"name"`              // e.g. "NVIDIA GeForce RTX 3090"
 	VRAMMb int    `yaml:"vram_mb,omitempty"` // e.g. 24576
 	Count  int    `yaml:"count,omitempty"`   // defaults to 1
 }
@@ -216,8 +216,18 @@ func hasService(manifest *CitadelManifest, serviceName string) bool {
 	return false
 }
 
-// addServiceToManifest adds a service to the manifest and writes it to disk.
+// addServiceToManifest adds a service to the manifest and writes it to disk,
+// honoring the hardcoded capability-tag map for embedded/catalog services.
 func addServiceToManifest(configDir, serviceName string) error {
+	return addServiceToManifestWithTags(configDir, serviceName, nil)
+}
+
+// addServiceToManifestWithTags adds a service to the manifest and writes it to
+// disk. In addition to the hardcoded capability-tag map (back-compat for
+// embedded/catalog services), it merges any module-declared routing tags
+// (service.yaml's node_tags) into Node.Tags, so third-party engines become
+// routable without a CLI change. Tags are deduped via containsTag.
+func addServiceToManifestWithTags(configDir, serviceName string, nodeTags []string) error {
 	manifestPath := filepath.Join(configDir, "citadel.yaml")
 
 	// Read existing manifest
@@ -237,7 +247,7 @@ func addServiceToManifest(configDir, serviceName string) error {
 		ComposeFile: filepath.Join("./services", serviceName+".yml"),
 	})
 
-	// Auto-add capability tags for specific services
+	// Auto-add capability tags for specific embedded/catalog services (back-compat).
 	serviceTags := map[string][]string{
 		"extraction": {"extraction:gliner2", "model:gliner2-base-v1"},
 	}
@@ -246,6 +256,13 @@ func addServiceToManifest(configDir, serviceName string) error {
 			if !containsTag(manifest.Node.Tags, tag) {
 				manifest.Node.Tags = append(manifest.Node.Tags, tag)
 			}
+		}
+	}
+
+	// Merge module-declared routing tags (service.yaml node_tags).
+	for _, tag := range nodeTags {
+		if tag != "" && !containsTag(manifest.Node.Tags, tag) {
+			manifest.Node.Tags = append(manifest.Node.Tags, tag)
 		}
 	}
 
--- a/cmd/module.go
+++ b/cmd/module.go
@@ -429,9 +429,14 @@ func buildModuleInstallCallbacks() controlcenter.ModuleInstallCallbacks {
 				}
 			}
 
-			_, configDir, err := findOrCreateManifest()
+			nodeManifest, configDir, err := findOrCreateManifest()
 			if err != nil {
 				return "", err
+			}
+			// Parity with the CLI: report an already-installed module cleanly
+			// instead of letting it surface as a confusing port conflict.
+			if hasService(nodeManifest, manifest.Name) {
+				return "", fmt.Errorf("module '%s' is already in the node manifest", manifest.Name)
 			}
 			servicesDir := filepath.Join(configDir, "services")
 
--- a/cmd/module.go
+++ b/cmd/module.go
@@ -128,14 +128,20 @@ func runModuleInstall(cmd *cobra.Command, args []string) error {
 
 	// Resolve the external source: clone/update the repo and load its manifest.
 	fmt.Printf("Resolving module source %s ...\n", color.CyanString(src.Raw))
-	manifest, composeSrc, err := catalog.ResolveSource(src)
+	resolved, err := catalog.ResolveSource(src)
 	if err != nil {
 		return err
 	}
+	manifest := resolved.Manifest
 
-	// Print a summary so the operator can confirm what will be installed.
-	image := composeImage(composeSrc)
-	printModuleSummary(src, manifest, composeSrc, image, overrides)
+	// Forward-compat: warn (don't fail) on a newer manifest schema.
+	if w := catalog.SchemaWarning(manifest); w != "" {
+		fmt.Println(color.YellowString("  Warning: " + w))
+	}
+
+	// Print a summary (incl. resolved commit + image) so the operator can confirm
+	// what will be installed -- a real TOFU/provenance prompt.
+	printModuleSummary(src, resolved, overrides)
 
 	if !moduleInstallYes {
 		fmt.Print("\nProceed with install? [y/N]: ")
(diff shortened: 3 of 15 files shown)


sunapi386:
Strong direction — decoupling service definitions from the CLI release cadence is the right call, and the self-describing `citadel/service.yaml` + `compose.yml` (module owns its image) is exactly the npm/helm-tap/homebrew model. A few recommendations for making it scale as a **community ecosystem** without giving up the "sovereign/secure" promise, roughly in priority order.

### 1. Trust & integrity (the load-bearing one)
Installing an arbitrary git repo = running an arbitrary container on a fabric node. Tag/branch trust isn't enough for a security-positioned product.
- **Close the deferred SHA-pin gap** and go further: resolve and record the **image digest (`sha256:…`)** at install, not just the git ref. Tags and `:latest` are mutable; a node's service set should be reproducible and tamper-evident.
- **Lockfile** (`ConfigDir()/modules.lock`) capturing `source@commit` + resolved image digest per installed module → reproducible nodes + drift detection.
- **TOFU prompt + provenance**: on first install of an untrusted source, show the resolved image and confirm (already non-interactive-friendly via overrides); persist `source URL + ref + commit` into the installed manifest so `module list` shows *where each service came from*. Later: optional cosign/sigstore signature or a verified-publisher allowlist.

### 2. Push capability tags INTO the module manifest
Today `cmd/manifest.go` hardcodes `serviceTags` per service name (`extraction`, etc.). That's the same coupling this PR removes for compose — it should move too. Let the module's `service.yaml` declare its routing tags (`tags: [engine:tei, task:embedding]`); the installer copies them onto the node manifest. Otherwise every community service still needs a CLI change to be routable. **This is the piece that makes third-party *engines* (not just images) actually work on the fabric.**

### 3. Reproducibility & versioning
- Support `@<sha>` (the deferred item) — mutable refs aren't acceptable for security-sensitive nodes.
- Semver/channel resolution later (`@^1.2`, `@stable`) like helm/npm.
- `module update` / `module list --outdated` / cache GC; auto-rollback to the previous pinned version if a freshly-installed module fails its `health_check`.

### 4. Preflight against node capabilities
`requires` (gpu/vram/arch) already exists — have the installer check it against detected capabilities and refuse/warn on mismatch. Concrete example from the embedding work: TEI images are CUDA-arch-specific (`86-*` = Ampere); a module that hardcodes one image will silently fail on a different GPU. So either modules declare `requires.arch` and the installer enforces it, or the image is parameterized and arch is resolved at install.

### 5. Conflicts & namespacing
Two modules wanting the same host port or `citadel-<name>` container will collide. Detect on install; namespace container names by source; keep `--set PORT=` overrides as the escape hatch.

### 6. Keep the curated set as the trust anchor
Don't position this as *replacing* the embedded/`citadel-services` catalog — keep that as the **standard library**: curated, offline-first, zero-extra-trust (vllm/ollama/tei). Module-sources are the **long tail**. One format (`service.yaml`), two trust tiers (curated vs community). A central index repo → `citadel module search` gives discovery without centralizing publication.

### Minor
- `schema_version` in `service.yaml` so the format can evolve without breaking old modules.
- Private-repo creds caveat is handled well; consider documenting a per-source credential hint in the error (which token/host).

Happy to dogfood once this lands — I have a first real **embedding module** (TEI / `gte-multilingual-base`, tracked in #343) lined up to install via this path, which will exercise the arch-preflight (#4) and tags-in-manifest (#2) cases specifically.

Issue #312, opened 2026-06-24
bug: 'citadel work' hangs forever on exit (SIGINT never terminates)


sunapi386:
## Problem

A locally-run `citadel work` (control center TUI) **hangs forever on exit** — Ctrl+C / SIGINT does not terminate the process; it has to be force-killed. Reported on macOS (2026-06-23).

```
^C   (nothing happens; process never exits)
```

This blocks clean restarts and leaves orphaned processes holding the VPN identity, the terminal server port, etc.

## Likely cause

Something in the shutdown path doesn't observe context cancellation / the signal, so the main goroutine blocks waiting on it. Candidates to audit (in `cmd/work.go` and the control-center lifecycle):

- A blocking `Read()`/`Accept()` on the embedded tsnet/VPN listener, terminal server (`internal/terminal`), or a PTY that isn't unblocked on shutdown (no deadline set, fd not closed).
- The periodic auto-updater goroutine (`internal/update.AutoUpdater`) or heartbeat publisher not stopping on context cancel.
- A `WaitGroup.Wait()` / channel receive in the shutdown sequence that waits on a goroutine which never returns.
- tview `Application.Stop()` not being called, or the signal handler not wired to it.

## Acceptance

- SIGINT/SIGTERM to `citadel work` causes a clean exit within a couple of seconds.
- All background goroutines (VPN, terminal server, heartbeat, auto-updater, console PTYs) are cancelled and joined; no orphaned process.
- Add a guard/timeout so shutdown can never block indefinitely (force-exit after a bounded grace period if a goroutine misbehaves).

Reproduced on macOS; investigate cross-platform.

PR #386, opened 2026-07-01
fix(work): start managed services asynchronously so job subscription isn't blocked (#384)

### LINKED ISSUE #384: [reliability] 'citadel work' blocks job-stream subscription on managed-service (vLLM) startup — head-of-line
## Problem

`citadel work` starts managed services from the manifest **before** it connects to the VPN and subscribes to its job streams. When a managed service is slow to start — e.g. `vllm` pulling an image / loading a model onto the GPU — the whole worker startup blocks on it, so the node stays **offline / unsubscribed** (and node-targeted jobs like `terminal_exec` / `cobrowse_*` time out) until the service finishes coming up.

Observed live on ubuntu-gpu: startup sat at
```
--- Starting 4 service(s) ---
   - Starting vllm...
   Container runtime: docker
```
for the entire observation window, and the worker only subscribed to its per-node stream once started with `--no-services`. As a workaround the node's systemd unit currently runs `citadel work --no-services`, which means it serves jobs but doesn't auto-start vLLM.

## Why it matters

Job serving (the node's core reason to exist) should not be gated on optional heavy services. A GPU node that takes minutes to load a model is unreachable for *all* job types during that window, and a service that hangs makes the node permanently unreachable even though the worker itself is healthy.

## Proposed direction

- Start managed services **asynchronously / in the background**, and connect + subscribe to job streams immediately (don't await service readiness before subscribing).
- Report per-service startup status via the existing status/pubsub channel rather than blocking the main path.
- Optionally a `--services-async` default with a `--wait-services` opt-in for callers that really want to block.

Surfaced while making co-browse durable under systemd (#380 / #4373).

sunapi386:
## Context

`citadel work` is the primary node command: it starts managed services from the manifest, connects to the VPN, and subscribes to its job streams so it can serve jobs (`terminal_exec`, `cobrowse_*`, code/file ops, etc.). Until now it did those **in order** — services first, subscription second — so a slow managed service blocked the whole worker startup.

Observed live on `ubuntu-gpu` (issue #384): a manifest with `vllm` sat at

```
--- Starting 4 service(s) ---
   - Starting vllm...
   Container runtime: docker
```

for the entire observation window while vLLM pulled its image / loaded the model onto the GPU. The node stayed **offline / unsubscribed** and served **no** jobs of any kind during that window. The workaround was to run the systemd unit with `--no-services` — which keeps the node reachable but never auto-starts vLLM. Job serving (the node's reason to exist) should not be gated on optional heavy services.

## Summary

Managed-service startup is now **asynchronous by default**. `runWork` kicks off service startup in a background goroutine and proceeds immediately to VPN connect + job-stream subscription, so the node comes online and serves jobs while services are still warming up.

**`cmd/work.go`**

- **Extracted** the old `autoStartServices()` loop into `startManagedServices(ctx) []startedService`:
  - Honors `ctx` cancellation — checks `ctx.Err()` between services so a slow startup goroutine stops launching the rest (and doesn't orphan) once shutdown is requested.
  - Returns the subset of services it **actually started** (skipping already-running containers), so shutdown only tears down what this worker launched — mirroring how the co-browse manager tracks only the Chromium process it spawned.
  - Per-service failures are logged as warnings and never abort the sequence; a heavy/failing service no longer blocks the others.
- **Async by default**: the startup call runs in a goroutine. `--wait-services` restores the old blocking behavior (subscribe only after every service is up). `--no-services` is unchanged.
- **Graceful shutdown preserved** — mirrors the existing co-browse teardown in the signal handler. Order matters: `cancel()` (so the startup goroutine stops launching more services) → **arm the #312 watchdog** (second-signal + grace-period force-exit) → `stopManagedServices(...)`. The watchdog is armed *before* teardown because teardown shells out to `docker compose down` (potentially ×N) and a wedged docker daemon must never hang the handler past the grace window and orphan the node's VPN identity. Like co-browse (kill the process, keep the persistent profile), services are stopped via `docker compose down` **without `-v`** (containers stop, model-cache volumes survive) or `StopNativeService`. A `defer` also runs teardown on normal exit; the second call is a no-op because the started-set is drained on first use. Access to the shared set is mutex-guarded since the signal handler can fire while the startup goroutine is still filling it in.
- **Grace window**: `shutdownGracePeriod` is 5s. A `docker compose down` across several services can exceed that (docker's default per-container stop timeout), in which case the (now-armed) watchdog force-exits mid-teardown — acceptable, since the containers are detached and the next start reconciles.

**Status reporting** is unaffected: the `internal/status` collector polls actual container state, so a slow or failed service simply reports as not-running regardless of start ordering — nothing is swallowed.

### Behavior change to confirm (called out for review)

Teardown-on-exit means a systemd restart now **stops** managed services (vLLM reloads its model into the GPU on the next start), whereas the old detached behavior let containers survive a restart. This is **not** an availability regression — the async fix keeps the node subscribed and serving jobs during that reload — but it is a deliberate semantic change (matches the "stop what we started" shutdown contract) and worth a reviewer sign-off. `--wait-services` is available for callers who want the old blocking sequence.

Known minor race (acceptable, containers are detached and the next start reconciles): if shutdown fires while a service is mid-`startService`, that single in-flight service may not be in the started-set yet and could be left running; the `ctx.Err()` check still prevents launching any *further* services.

## Test plan

| Test | Coverage |
| --- | --- |
| `TestStartManagedServicesNoManifest` | No manifest → returns immediately, nothing started (never blocks/fails the worker) |
| `TestStartManagedServicesCanceledContextDoesNotStart` | **Core #384 contract**: a canceled context (shutdown-during-startup) halts before shelling out to docker — subscription is not gated on service readiness |
| `TestStopManagedServicesEmptyIsNoop` | Empty teardown is a no-op (double-call path: signal handler + deferred cleanup) |
| `TestWaitServicesFlagDefaultsToAsync` | `--wait-services` registered, defaults to `false` (async is the default); `--no-services` still registered |

Verification run:

- `gofmt -l` — clean
- `go build ./...` — passes
- `go vet ./cmd/... ./internal/...` — passes
- `go test ./cmd/...` — passes
- `go test ./internal/...` — passes except `internal/status/TestServerStartAndShutdown`, which fails only because the **live** `citadel` worker on this host holds `:8080` (`listen tcp :8080: bind: address already in use`). Environmental, unrelated to this change (this PR does not touch `internal/status`).

**Live-node acceptance check (why this is a draft):** on a node with services enabled (drop the `--no-services` workaround from the systemd unit), start `citadel work` and confirm the node **subscribes to its job streams and serves jobs immediately** while vLLM is still loading the model — i.e. `citadel_node_info` / a `terminal_exec` succeeds during the vLLM warm-up window, instead of timing out until the model finishes loading. Also confirm SIGTERM stops the managed services cleanly.

## Related

- Closes #384
- Surfaced while making co-browse durable under systemd (#380 / #4373)

### DIFF
--- a/cmd/work.go
+++ b/cmd/work.go
@@ -203,11 +203,14 @@ func runWork(cmd *cobra.Command, args []string) {
 			fmt.Fprintf(os.Stderr, "   - Warning: co-browse browser stop: %v\n", err)
 		}
 		// Cancel first so the async service-startup goroutine stops launching any
-		// further services, then tear down the ones we already started. Like
-		// co-browse, this stops the containers/processes we spawned but preserves
-		// their data (no `-v`), and only touches services this worker launched.
+		// further services.
 		cancel()
-		teardownStartedServices()
+		// Arm the watchdog BEFORE tearing down managed services. Service teardown
+		// shells out to `docker compose down` (potentially several times) and a
+		// wedged docker daemon could hang this handler; the watchdog (second
+		// signal + grace-period force-exit, issue #312) must already be running so
+		// a stuck teardown can never leave an orphaned node holding the VPN
+		// identity or the terminal-server port.
 		go func() {
 			select {
 			case <-sigs:
@@ -218,6 +221,10 @@ func runWork(cmd *cobra.Command, args []string) {
 				os.Exit(1)
 			}
 		}()
+		// Tear down the services we started. Like co-browse, this stops the
+		// containers/processes we spawned but preserves their data (no `-v`), and
+		// only touches services this worker launched.
+		teardownStartedServices()
 	}()
 
 	// Note: Update check is now handled by root.go's PersistentPreRun
--- a/cmd/work.go
+++ b/cmd/work.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -76,7 +77,8 @@ var (
 	workTerminalDebug bool
 
 	// Service auto-start flags
-	workNoServices bool
+	workNoServices   bool
+	workWaitServices bool
 
 	// Update check flag
 	workNoUpdate bool
@@ -149,6 +151,31 @@ func runWork(cmd *cobra.Command, args []string) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// Managed services this worker started (populated by the async startup
+	// goroutine below, read by the shutdown hook). Guarded by a mutex because
+	// the signal handler may fire while the startup goroutine is still filling
+	// it in — we only ever want to tear down services we actually launched.
+	var (
+		startedServicesMu sync.Mutex
+		startedServices   []startedService
+	)
+	recordStartedServices := func(svcs []startedService) {
+		startedServicesMu.Lock()
+		startedServices = append(startedServices, svcs...)
+		startedServicesMu.Unlock()
+	}
+	teardownStartedServices := func() {
+		startedServicesMu.Lock()
+		svcs := startedServices
+		startedServices = nil
+		startedServicesMu.Unlock()
+		stopManagedServices(svcs)
+	}
+	// Tear down managed services on any exit path (signal handler below also
+	// invokes this before the watchdog fires; the second call is a no-op because
+	// the slice is drained on first use).
+	defer teardownStartedServices()
+
 	// Apply --no-gateway / --no-terminal overrides (take precedence over defaults)
 	if workNoGateway {
 		workGateway = false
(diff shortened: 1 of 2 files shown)


sunapi386:
## Live acceptance test: PASS (on ubuntu-gpu / node 1054, RTX 3090)

Ran the `#386` binary with **services enabled** (vLLM `Qwen/Qwen3-8B`) and timed subscription vs. service startup:

- **Launched** 08:13:21 → log shows `--- Starting 4 service(s) --- / Starting vllm...` **and** `Per-node shell stream: ...:node:1054` — the worker proceeded to gateway/VPN/subscription instead of blocking on vLLM.
- **Subscribed** 08:13:52 (~31s — that's tsnet VPN connect, not vLLM).
- **Served a real job** at 08:14:30: `terminal_exec` → executed on `ubuntu-gpu-1nx1ud26` — **while vLLM was still loading** (no `citadel-vllm` container up, `/v1/models` not ready).

Under the pre-#386 behavior, `Starting vllm...` blocked subscription entirely (observed earlier: the worker sat there indefinitely, which is why the node's systemd unit needed the `--no-services` workaround). With #386, job serving is no longer gated on managed-service readiness. Acceptance criteria met.

Issue #388, opened 2026-07-01
TUI: make the Control Center mouse-clickable (feel like a GUI), starting with the Chat tab


sunapi386:
## Goal

Make the Control Center TUI **mouse-clickable** so it feels like a GUI, not a keyboard puzzle. Right now every interaction is a keybinding (Tab to switch panes, `0-3` for actions, Enter for details, arrows to move). New users have to learn the map before they can do anything. If you can just **click a tab, click a peer, click Send**, the TUI instantly feels like a desktop app — same slick, "nothing to memorize" feel as [TORLINK](https://github.com/baairon/torlink).

This pairs naturally with the flicker-free fullscreen renderer: fullscreen TUIs already turn on terminal mouse reporting, so the events are right there — we just aren't listening for them.

Start with the **Chat tab** (actively being built) since a chat UI is where click-to-focus, click-a-channel, click-a-peer, scroll-the-history feels most like a real app.

## Current state

- Control Center is **tview** (`internal/tui/controlcenter/controlcenter.go:610`, `tview.NewApplication()`); dashboard also tview (`internal/tui/dashboard/tview.go:42`).
- **Mouse is not enabled anywhere** — no `EnableMouse` / `SetMouseCapture` calls in `internal/`. tview defaults to mouse off.
- Navigation is 100% key-based: `KeyTab` to cycle panes, `0-3` action hotkeys, Enter for details (`controlcenter.go:872`, help bar at `:1961`).
- Chat page (`internal/tui/controlcenter/chat_page.go`) is a Slack-style layout: `channelBox` + `peersBox` sidebar, `msgView` message pane, `statusBar`, `input` InputField — all navigable only by keys today.
- bubbletea sub-views (`repl`, `dashboard/status`, `whimsy`) create programs **without** `tea.WithMouseCellMotion()`.

## The fix (small surface, big UX win)

tview has first-class mouse support built in; primitives (`Table`, `List`, `TextView`, `InputField`, `Button`, `Pages`) already handle clicks/scroll internally once the app opts in.

1. **Turn it on:** `cc.app.EnableMouse(true)` on the Control Center app (and the dashboard app).
2. **Wire click intent to existing actions** — reuse the handlers the keybindings already call:
   - Click a pane/panel → focus it (same as `Tab` landing there).
   - Click a list row (Services, Peers, Jobs, chat channels/peers) → select it; double-click → the `Enter` "details" action.
[...]

PR #393, opened 2026-07-01
feat(tui): smooth Tab/arrow navigation + unify modules (WhatsApp/Proxmox) + seed sources (#388 follow-up)

sunapi386:
## Context

Jason interactively tested the Control Center after #388/#390 and flagged four IA/navigation issues: navigation felt clunky (too many `alt+N` hotkeys to memorize), the Install-module page trapped focus and forced blind typing, WhatsApp had a redundant top-level tab, and Proxmox appeared as a top-level tab with no obvious reason. This PR makes the Control Center keyboard-first ("nothing to memorize" — the TORLINK feel from #388) and unifies module management.

Everything is reachable with **arrow keys + Tab**. `alt+N` stays as an optional accelerator; the mouse support from #390 is untouched.

### 1. Smooth Tab/arrow navigation (primary), `alt+N` kept as accelerators

The load-bearing decision: **bubble-after-delegate**. `PageManager.HandleGlobalInput` delegates each key to the active page first, then bubbles an *unconsumed* `Tab`/`Shift+Tab` up to tab-switching. A page consumes Tab for its own intra-page focus movement (dashboard pane cycling, module list ↔ form) and returns it **unconsumed only at its navigation boundary** to hand off to the next/previous tab.

- **This makes the default correct**: a page that does nothing with Tab (e.g. `chat_page.go`, `settings_page.go`) gets tab-switching for free — which is exactly the "convention parallel agents rely on".
- **Dashboard** pane cycling now bubbles at the edges (Tab past the last pane advances to the next tab; Shift+Tab before the first pane falls back to the previous tab) instead of wrapping forever inside the dashboard.
- **`Escape` is not captured globally** — the dashboard's `Esc = quit-confirm` keybinding is preserved. "Escape defocuses an input back to nav" is a per-page rule (implemented in the module page).
- Help modal + help bar copy updated to teach Tab-first nav.

### 2. Install Module (`alt+6`): fix intra-page nav + seed approved sources

- **Seeded "Known modules" list**: a new `ListSources` callback pre-populates a selectable list from the catalog registry + curated module index (`moduleSourcesFromCatalog`), fail-soft to a small curated set (`vllm`, `ollama`, `comfyui`) when the catalog isn't cloned. The user picks a module instead of typing into a blank input blind.
- **Fixed the focus bug**: Tab now cycles **list → form fields → tabs**. The form *owns* Tab (a `SetInputCapture` handler that returned the event to let tview cycle fields natively would be intercepted by the bubble check first), advancing form focus itself via `nextFormFocus` and returning the event only at the true first/last boundary. `Escape` from the form returns focus to the list so an input never traps the user.

### 3. WhatsApp (`alt+5`) folded into the module list

WhatsApp is a community module, so its standalone tab is removed: the page is registered **hidden** and surfaced as a "WhatsApp (community module)" row in the module list that switches to it (via a `SelectSpecial` hook). All bridge functionality (deploy / stop / pairing QR) is preserved unchanged — only the entry point moved.

### 4. Proxmox (`alt+4`) — decision: **keep gate-on-detection, do NOT fold**

Investigation showed Proxmox is **already gated on real detection**: `buildProxmoxConfig()` returns `Enabled: true` only when saved Proxmox config exists **or** a local host is detected (`/etc/pve` + `pvesh` via `DetectProxmox()`), and `Run()` only registers the tab when enabled. So the "shows unconditionally" premise was false — it appeared for Jason because he has saved config or a detected host (that answers his "how is it detected?" question).

**Rationale for keeping the gate rather than folding:** Proxmox is host-infrastructure management, not an installable module. Folding it into the module list would misrepresent it. The gate is now expressed through a pure, testable `proxmoxTabVisible()` seam (also defends against an `Enabled`-but-empty-URL config).

## Test plan

| Test group | Coverage |
|---|---|
| `nextVisibleIndex` / `prevVisibleIndex` / round-trip | Tab-cycle order: skip hidden, wrap at ends, single/empty no-op |
| `nextFormFocus` | Form boundary: cycle fields/buttons, hand off only at first/last (the reported bug) |
| `moduleSourcesFromCatalog` | Seeding: registry trusted, index untrusted, dedupe (registry wins), empty → fallback |
| `proxmoxTabVisible` | Gate: shown only when detected (Enabled + BaseURL) |
| module-page (headless `tview.Application`) | List seeding, trusted-check prefix, empty placeholder, special-row switch, normal-row seed |

```
gofmt -l          → clean
go build ./...    → ok
go vet ./internal/tui/... ./cmd/...  → ok
go test ./internal/tui/... ./cmd/... → ok
```

`internal/status` `TestServerStartAndShutdown` fails with `bind: :8080 address already in use` — the known environmental flake (a live citadel node runs on the test host), unrelated to this change.

## Notes for parallel agents (shared convention)

Pages with **internal focusable elements** (multi-field forms) must own Tab and return the event only at their first/last boundary (see `ModulePage.HandleInput`). Pages whose controls are activated by single keys (`settings_page.go`, `whatsapp_page.go`) can ignore Tab and let it bubble — the default is correct for them. Tab registrations for `chat_page.go` / `settings_page.go` were left intact; only the tab list/order was adjusted (WhatsApp hidden).

**This is a draft** pending interactive verification of the live TUI (arrow/Tab feel, module picker, WhatsApp fold) on a real terminal.

## Related

- Follows #388 (mouse-clickable Control Center) and #390.
- Addresses Jason's post-#388 interactive feedback (all four IA/nav issues).

### DIFF
--- a/cmd/module.go
+++ b/cmd/module.go
@@ -613,12 +613,78 @@ func confirmYes() bool {
 	return resp == "y" || resp == "yes"
 }
 
+// fallbackModuleSources is the small curated set surfaced when the catalog has
+// not been cloned yet, so the TUI's module picker is never blank on a fresh node.
+// These are first-party catalog names (Tier 0, trusted) resolvable once the
+// catalog is fetched; the resolve step clones on demand.
+var fallbackModuleSources = []controlcenter.ModuleSource{
+	{Name: "vllm", Source: "vllm", Description: "vLLM GPU inference server", Trusted: true},
+	{Name: "ollama", Source: "ollama", Description: "Ollama local model runner", Trusted: true},
+	{Name: "comfyui", Source: "comfyui", Description: "ComfyUI image generation", Trusted: true},
+}
+
+// moduleSourcesFromCatalog maps the loaded catalog registry + curated module
+// index into the TUI's seeded picker rows. Pure over its inputs (no IO) so the
+// seeding logic is unit-testable. Catalog registry entries are first-party
+// (Tier 0) and marked trusted; index entries carry their source repo and are
+// left untrusted unless the index says otherwise (it currently doesn't, so they
+// render with the untrusted warning at resolve time). Duplicate names (registry
+// vs index) are de-duplicated, registry winning. When both inputs are empty the
+// caller substitutes the fallback set.
+func moduleSourcesFromCatalog(reg *catalog.Registry, idx *catalog.ModuleIndex) []controlcenter.ModuleSource {
+	var out []controlcenter.ModuleSource
+	seen := make(map[string]bool)
+	if reg != nil {
+		for _, e := range reg.Services {
+			if e.Name == "" || seen[e.Name] {
+				continue
+			}
+			seen[e.Name] = true
+			out = append(out, controlcenter.ModuleSource{
+				Name:        e.Name,
+				Source:      e.Name, // catalog name resolves via the central catalog
+				Description: e.Description,
+				Trusted:     true,
+			})
+		}
+	}
+	if idx != nil {
+		for _, e := range idx.Modules {
+			if e.Name == "" || seen[e.Name] {
+				continue
+			}
+			seen[e.Name] = true
+			out = append(out, controlcenter.ModuleSource{
+				Name:        e.Name,
+				Source:      e.Source, // owner/repo or git URL from the curated index
+				Description: e.Description,
+				Trusted:     false,
+			})
+		}
+	}
+	return out
+}
+
+// listModuleSources loads the catalog registry + curated index and maps them into
+// the TUI picker rows. Fail-soft: any load error (catalog not cloned, no index
+// published) falls back to the small curated set so the picker is never blank.
+func listModuleSources() []controlcenter.ModuleSource {
+	reg, _ := catalog.LoadRegistry()
+	idx, _ := catalog.LoadModuleIndex()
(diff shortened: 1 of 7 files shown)


sunapi386:
### Interactive verification checklist (why this PR is a draft)

The unit tests cover the pure index/seeding logic; these behaviors need a real terminal to confirm the *feel*:

- [ ] Dashboard: Tab moves through the 7 panes, then on to the next tab; Shift+Tab from the first pane jumps to the last tab. Arrow keys move within a pane.
- [ ] `alt+1..N` still jump straight to a tab; mouse clicks on tabs/rows still work (#390 unregressed).
- [ ] Install Module: the "Known modules" list is pre-populated; Enter on a row seeds the form; Tab cycles **list → form fields → tabs**; Escape from the form returns to the list (no focus trap).
- [ ] WhatsApp: no standalone tab; the "WhatsApp (community module)" row in the module list opens the bridge page; deploy/stop/QR all work; Tab returns to the tab bar.

Two known, navigable (never-stuck) quirks of folding WhatsApp into a hidden page — confirm they read as acceptable, not as glitches:
- On the WhatsApp page, **no tab is highlighted** in the bar (the active page is hidden, so nothing is lit). The in-page "Community module — press Tab to return" hint mitigates this.
- **Shift+Tab from WhatsApp lands on Console**, not the module page (previous-visible-tab math), while forward **Tab correctly lands on the module page**. Both directions always reach a visible tab.

---
*Generated by Claude*

Issue #415, opened 2026-07-01
diffusers host port 8102 collides with TEI + SERVICE_START container has no published ports (endpoint unreachable)


sunapi386:
## Context
Found during live prod verification of the diffusers deploy loop on node 1054 (RTX 3090). After the #413 manifest-reconcile fix let the node start diffusers, the deploy still could not expose the service.

## Two related bugs
1. **Host port 8102 collides with TEI.** The embedded `services/compose/diffusers.yml` maps host `8102:7860`. But TEI already owns host `8102` on a GPU node (the gateway routes `/v1/embeddings -> 127.0.0.1:8102`). `docker compose up` fails: `Bind for 0.0.0.0:8102 failed: port is already allocated`. The 8100/8101/8102 host-port sequence in the compose comment did not account for TEI.
2. **SERVICE_START container had no published ports at all.** The container created by the SERVICE_START handler came up with `NetworkSettings.Ports == {}` (verified via `docker inspect`), so even without the collision the provisioned endpoint (host:8102) would be unreachable. The server was healthy on the internal contract port 7860, but nothing was bound on the host.

## Impact
A provisioned diffusers endpoint is unreachable. In the live test I had to run the container manually on a free host port (8188) to reach `/generate`.

## Fix
- Give diffusers a non-colliding host port (pick one not used by TEI/terminal/vllm/transcribe; audit the full host-port map), or allocate dynamically and report the chosen port back.
- Ensure the SERVICE_START path actually applies the compose port mapping to the running container, and returns the reachable host endpoint.
- Add a test asserting the started container has the expected host port bound.

## Related
citadel-cli #413/#414 (manifest reconcile), aceteam #4472, #4468.

---
*Generated by Claude Opus 4.8*

sunapi386:
Bug (1) — diffusers host 8102 colliding with TEI — is fixed in **#410** (citadel-owned host-port registry: diffusers moves to a non-colliding host port, and a collision-guard test now fails CI if any two services/apps claim the same host port). This issue's remaining scope is **bug (2)**: the SERVICE_START handler bringing the container up with `NetworkSettings.Ports == {}` (no host binding), so even post-collision the endpoint is unreachable. A PR for (2) is incoming (grouped with the #413 manifest-reconcile fix, since both block the same diffusers deploy loop).

PR #419, opened 2026-07-01
fix(services): reconcile manifest with embedded services on upgrade + publish SERVICE_START host ports (#413, #415)

### LINKED ISSUE #413: Upgrading the binary does not add newly-embedded services to citadel.yaml (deploys fail 'not found in manifest')
## Context

Discovered during a live prod test of the Fabric Model Library phase-1 deploy loop (epic aceteam-ai/aceteam#4472). Node `1054` (RTX 3090, `ubuntu-gpu`) is on **v2.55.0**, which embeds the diffusers service in the `ServiceMap` and advertises it via `GetAvailableServices()`. The platform correctly resolved `sd-3.5-medium` to engine `diffusers` and dispatched a deploy to the node. It failed at the node:

```
service "diffusers" not found in manifest (known: vllm, ollama, llamacpp, tei)
```

No image pulled, no container.

## Root cause

Capability advertisement and runtime runnability are **decoupled**:

- `services.GetAvailableServices()` (embedded `ServiceMap`) advertises `diffusers` in the heartbeat `available_services` (correct, wired in `internal/status/collector.go:129` `populateServices`).
- The node's runtime manifest `~/citadel-node/citadel.yaml` was written at `init` time (pre-diffusers) and lists only `vllm, ollama, llamacpp, tei`.
- A binary upgrade (`citadel update` -> v2.55.0) does **not** reconcile `citadel.yaml` with the newly-embedded `ServiceMap`, so `SERVICE_START diffusers` is rejected with "not found in manifest."

Net: a node advertises a capability it cannot actually start. Nothing in upgrade, init, or the SERVICE_START handler bridges the gap.

## Proposed fix (pick one, discuss)

1. **Reconcile on upgrade/startup** (preferred): on `citadel update` and/or agent start, add a service block to `citadel.yaml` for every embedded `ServiceMap` entry missing from the manifest (idempotent, additive, never removes operator entries). This keeps advertised == runnable.
2. **SERVICE_START fallback to embedded compose**: if a requested service is absent from the manifest but present in the embedded `ServiceMap`, start it from the embedded compose (and persist the block).
3. **`citadel service enable <name>`** explicit command + have the platform deploy call it before start.

Option 1 or 2 closes the loop without operator action. Whichever we choose, add a test asserting a node upgraded from a pre-diffusers manifest can start `diffusers`.

## Repro

1. Node with a pre-diffusers `citadel.yaml` (services: vllm/ollama/llamacpp/tei), binary upgraded to v2.55.0.
2. From the platform: `provision_model("sd-3.5-medium")` -> schedules to the node, dispatches diffusers.
3. `service_status diffusers` -> "not found in manifest".

## Related

- aceteam-ai/aceteam#4472 (Fabric Model Library epic), citadel-cli #404 (diffusers service), #408 (self-declare `available_services`).
- Companion platform-side issue: deploy should verify runnability + pre-pull image before SERVICE_START.

---

sunapi386:
## Context

The diffusers deploy loop was dead in two coupled places, and a node could advertise a serving capability it was physically unable to start.

When citadel embeds a **new** service (diffusers landed in `services.ServiceMap` in v2.55.0) it immediately advertises it in the heartbeat via `services.GetAvailableServices()` / `internal/status/collector.go`'s `populateServices`. But the node's on-disk runtime manifest (`~/citadel-node/citadel.yaml`) was written at `init` time, *before* that service existed — and a plain `citadel update` never touches the manifest. So the fabric schedules a `SERVICE_START diffusers` to a node that reports it can run it, and the node rejects it:

```
service "diffusers" not found in manifest (known: vllm, ollama, llamacpp, tei)
```

Separately, even once a service *was* startable, the container it produced came up with `NetworkSettings.Ports == {}` (verified via `docker inspect`) — the compose's `ports:` mapping was silently not applied, so a perfectly healthy server on its internal contract port was unreachable on the host.

Net effect: advertise → schedule → reject, and when it didn't reject, → unreachable. This PR closes both halves so a fresh binary self-heals its manifest and actually publishes the port its compose declares.

### Bug A — manifest reconcile (`Closes #413`)

- **Startup reconcile (primary, Option 1).** New `reconcileManifestServices()` in `cmd/manifest.go`: for every embedded `ServiceMap` entry missing from the manifest, it adds an additive service block and materializes the embedded compose into `services/<name>.yml`. It is **additive-only** (never removes/overwrites operator-defined entries), **idempotent**, preserves unknown operator entries (e.g. a hand-added `tei`) and all non-service manifest content (`node`, `capabilities`).
- **Ordering matters.** Wired into `runWork` **after** `startManagedServices`, on purpose. `startManagedServices` boots *every* manifest entry, and #384's design is "never gate startup on heavy GPU services." If reconcile ran before it, a node whose manifest just gained diffusers/sglang/lmstudio would kick off unrequested multi-GB image pulls + model downloads at every boot. Running reconcile after means the reconciled entries only exist so an **on-demand** `SERVICE_START` can find them — they never launch unbidden.
- **On-demand self-heal (Option 2, belt-and-suspenders).** In the `SERVICE_START` handler, if a requested service is absent from the manifest but present in the embedded `ServiceMap`, it materializes the compose + persists an additive manifest block and proceeds. This also covers the TUI-launched worker path (`ccStartWorker`/`runTUIWorker`), which doesn't run the startup reconcile — so a `SERVICE_START` to a TUI worker self-heals too. Guarded by a path-traversal-safe service-name check.

> **Deviation note:** the task named the startup reconcile as the primary fix. I kept it as primary, but the on-demand materialize in the handler is what makes both worker boot paths robust without the boot-time auto-start regression. Both ship.

### Bug B — publish `SERVICE_START` host ports (`addresses the SERVICE_START-no-published-ports half of #415`; the 8102↔TEI collision half is fixed in #410)

- **Force recreation.** The docker `SERVICE_START` path now passes `--force-recreate` to `docker compose up -d`. The `Ports == {}` symptom happens when a stale `citadel-<svc>` container from a prior run (with a different or absent port mapping) is reused instead of recreated from the current compose. `cmd/service.go`'s startup path already `rm -f`'d stale containers; the handler path did not. `--force-recreate` guarantees the container is built from *this* compose definition. We only reach this branch when the service isn't already running, so recreating is safe.
- **Report the endpoint.** `serviceResult` gains `host_ports`, populated on a successful docker start, so the caller knows where the service is reachable on the node.
- **`compose.HostPorts()`** — a new pure, daemon-free parser (`internal/compose/modifier.go`) for the published host ports in a compose file (short form `HOST:CONTAINER`, `IP:HOST:CONTAINER`, `/proto` suffixes, and long-form `published/target`; bare/ephemeral publishes skipped). Lets us report the endpoint and assert the intended publish in tests without a live docker.

**No host-port numbers touched.** This PR does not modify port *numbers* or `services/ports.go`/`services/compose/*.yml` host ports — that's #410's territory (see Related).

## Test plan

| Group | Coverage |
|-------|----------|
| `cmd` manifest reconcile | pre-diffusers manifest reconciles to include diffusers + all embedded services; additive (legacy + non-embedded `tei` survive); node identity preserved; compose materialized; idempotent (2nd run adds nothing, no dupes); operator-authored compose file never overwritten |
| `internal/jobs` materialize | on-demand materialize adds diffusers to a pre-diffusers manifest and passes the "known service" check; preserves `node`/`capabilities`; idempotent; unknown service → normal "not found" error; path-traversal name rejected |
| `internal/jobs` port publish | materialized diffusers compose reports its declared host port (asserts the intended publish without docker); `--force-recreate` present in the docker start path |
| `internal/compose` HostPorts | short/long form, `IP:HOST:CONTAINER`, `/proto` suffix, bare/ephemeral skipped, multi-service dedupe, invalid YAML → nil |

`gofmt -w .`, `go build ./...`, `go vet ./...`, and `go test ./internal/jobs/... ./internal/status/... ./internal/compose/... ./services/... ./cmd/...` all green.

## Related

- Closes #413 (manifest not reconciled with embedded services on upgrade)
- Addresses the "SERVICE_START comes up with no published ports" half of #415. The diffusers host-port **collision** (8102 vs TEI) is fixed separately in #410 (a citadel-owned host-port registry). No overlap: this PR changes *whether* the declared port is published, not *which* port it is.

### DIFF
--- a/cmd/manifest.go
+++ b/cmd/manifest.go
@@ -2,6 +2,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -268,6 +269,79 @@ func addServiceToManifestWithTags(configDir, serviceName string, nodeTags []stri
 
 	// Write back
 	return writeManifest(manifestPath, manifest)
+}
+
+// reconcileManifestServices makes the on-disk manifest consistent with the
+// services this binary can deploy. For every service embedded in the binary
+// (services.ServiceMap) that is missing from the manifest's services list, it
+// adds an additive service block and materializes the embedded compose file
+// into <configDir>/services/<name>.yml.
+//
+// This closes the "advertised-but-unstartable" gap (citadel-cli#413): when a
+// newer binary embeds a service (e.g. diffusers in v2.55.0) and advertises it in
+// the heartbeat, but the node's citadel.yaml was written at init time before that
+// service existed, a SERVICE_START would be rejected with "service not found in
+// manifest". A binary upgrade does not otherwise touch the manifest, so the node
+// keeps advertising a capability it cannot start until reconciled here.
+//
+// Guarantees:
+//   - Additive-only: never removes or overwrites operator-defined entries. An
+//     existing service (matched by name) is left untouched, including
+//     operator-authored compose files (ensureComposeFile is a no-op when the file
+//     already exists).
+//   - Preserves unknown entries: services in the manifest that are not embedded
+//     (e.g. a hand-added "tei") are retained as-is.
+//   - Idempotent: a second call is a no-op.
+//
+// It returns the names of services newly added (for logging). It never returns an
+// error that should abort startup; failures for a single service are collected
+// and surfaced, but the manifest write only includes services that materialized
+// cleanly.
+//
+// IMPORTANT ordering note for callers: this must run AFTER startManagedServices
+// so newly-reconciled embedded services are not auto-started on this boot. The
+// manifest lists only operator intent for the purposes of auto-start (#384: never
+// gate boot on heavy GPU services); reconciled entries exist so an on-demand
+// SERVICE_START can find them, not so they launch unbidden.
+func reconcileManifestServices(configDir string) (added []string, err error) {
+	manifestPath := filepath.Join(configDir, "citadel.yaml")
+
+	manifest, _, mErr := findAndReadManifest()
+	if mErr != nil {
+		return nil, fmt.Errorf("failed to read manifest for reconcile: %w", mErr)
+	}
+
+	// Iterate embedded services in a stable order so logs and writes are
+	// deterministic.
+	changed := false
+	for _, svcName := range services.GetAvailableServices() {
+		if hasService(manifest, svcName) {
+			continue // operator/prior-reconcile entry -- leave untouched
+		}
+
+		// Materialize the embedded compose file (no-op if it already exists).
+		if cErr := ensureComposeFile(configDir, svcName); cErr != nil {
+			err = errors.Join(err, fmt.Errorf("reconcile %s: %w", svcName, cErr))
+			continue
+		}
+
+		manifest.Services = append(manifest.Services, Service{
+			Name:        svcName,
+			ComposeFile: filepath.Join("./services", svcName+".yml"),
+		})
+		added = append(added, svcName)
+		changed = true
+	}
+
+	if !changed {
+		return added, err
+	}
+
+	if wErr := writeManifest(manifestPath, manifest); wErr != nil {
+		return added, errors.Join(err, fmt.Errorf("failed to write reconciled manifest: %w", wErr))
+	}
+
+	return added, err
 }
 
 // containsTag checks if a tag is already in the tags slice.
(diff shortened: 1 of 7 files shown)


sunapi386:
Closing as duplicate — while this was in flight, a parallel session already landed the #413/#415 fixes on main (e67486b + 15a5bbf reconcile embedded services into the manifest on demand; db32cc2 frees the diffusers host port + publishes SERVICE_START ports). Both issues are already CLOSED. This branch's independent implementation (materialize-on-demand + --force-recreate + compose.HostPorts parser) is redundant with what merged. Superseded.

Issue #383, opened 2026-06-30
[bug] 'citadel login --authkey' creates a DUPLICATE node instead of reattaching; reconnect fails when device token lacks 'write' scope


sunapi386:
### 1. Device-token scope failure blocks auto-recovery (silent)
On startup with stale tsnet state, `citadel work` tries to fetch a fresh authkey and fails:
```
VPN auto-recovery failed: could not fetch fresh authkey: device API token rejected (HTTP 403)
— Access denied: Missing required scope 'write' for path /api/fabric/authkey/generate
```
`citadel reconnect` hits the same 403 (it also fetches via the device token). The node stays disconnected → Headscale ID never resolves → per-node stream never subscribes → node-targeted jobs time out or leak to a peer. This is currently a **buried warning**; it should be a hard, actionable error (and old device tokens should be healable to the `write` scope without a full re-init).

### 2. `citadel login --authkey` creates a duplicate node
Working around (1) by minting an authkey and running `citadel login --authkey <key>` **created a new Headscale registration** (`ubuntu-gpu-1nx1ud26`, new IP) instead of reattaching to the existing node — because the fresh tsnet identity got a new node key. Result: a duplicate, and the original fabric node id was orphaned.

## Proposed direction

- `citadel login --authkey` / `reconnect` should **preserve the existing node identity** (reuse the stored machine key / reattach) rather than minting a new registration, or at minimum detect the same hostname and offer to replace the stale node.
- Surface the device-token scope failure as a clear, actionable error with the exact remediation (re-issue the key with `write`, or the heal path).
- Relates to the aceteam device-API-key scope UX.

sunapi386:
## Root-cause investigation: why `login --authkey` duplicated the node, and how to make it structurally impossible

Traced through the code. The duplicate required **two independent conditions** — the suffixed hostname (`ubuntu-gpu-1nx1ud26`) is dispositive, because Headscale keys nodes by **machine key**, not hostname: had tsnet reused the existing machine key it would have reattached with no new node and no suffix.

### Condition A — the stale-node reclaim failed *silently*
`login --authkey` already tries to clean up: `runNonInteractiveLogin` → `reclaimStaleNodeByHostname` (`cmd/login.go:78-80`) → `nexus.ReclaimStaleNode` → `POST /api/fabric/device-auth/deregister` (`internal/nexus/reclaim.go`). This fails because:
[...]

Issue #175, opened 2026-06-15
fix: VPN reconnect fails with stale tsnet state


sunapi386:
When running `citadel work --gateway --terminal`, VPN reconnect via `network.VerifyOrReconnect()` times out after 60 seconds when tsnet state is stale. This happens after reboots or when the Headscale preauth key expires. The node needs to re-pair to get a fresh auth key.

Current workaround: manually clear state (`rm tailscaled.state`) and re-login (`citadel login --authkey <key>`).

Proposed fix: detect stale state (e.g., connection timeout), clear the tsnet state directory, and trigger the device auth flow to get a fresh Headscale key automatically.

Related: Previously filed as #168, but that was about the sudo HOME mismatch. This is the deeper issue.

---
*Generated by Claude*

PR #435, opened 2026-07-02
feat(node): renew Headscale node key while online + harden no-authkey reconnect (#4583)

sunapi386:
## Context

Citadel side of the **P0 durable-key baseline** for the self-healing fabric node identity epic (**#4583**), sibling to backend PR **#4584**.

We hit this live: a node dropped off the VPN mesh on restart and could not rejoin without an admin. The failure mode has two halves:

1. **Registration ephemerality** (backend, #4584) — re-authed nodes were minted *ephemeral* Headscale keys, so an offline node vanished from Headscale and had to re-register on restart. #4584 makes registration keys non-ephemeral so a persistent node stays registered across an outage and the *existing* no-authkey reconnect re-establishes the session.
2. **Node-key expiry** (this PR) — even a persistent node's Headscale preauth-derived **node key can EXPIRE** over a long-lived session. **tsnet with a preauth AuthKey does not proactively renew node keys** — it only re-authorizes when `Start()` is called with a fresh authkey (interactive login is useless headless). So a healthy, online, long-running node can silently lapse its key, drop off the mesh, and then dead-end in the offline recovery path — which historically hit the `device_authkey:write` **403 landmine**.

The defense: while the node is healthy and online, notice the approaching expiry and re-authorize **in place** using the node's own already-authenticated session, so the key never lapses and the offline mint path is never reached.

## tsnet: what it does vs. what this adds

| | Behavior |
|---|---|
| **tsnet auto-renew?** | No. A preauth `AuthKey` authorizes once; the node key expires per Headscale policy and tsnet does not refresh it on its own. |
| **This PR's renewal** | In-place reauth via `LocalClient.Start(ipn.Options{AuthKey: fresh})` — the same mechanism as `tailscale up --auth-key`. Applies a fresh authkey to the **running** state machine, re-authorizing the node key **without tearing down the connection**. The machine key (and thus the node's IP) and all VPN listeners (`status`/`terminal`/`gateway`) bound at `citadel work` startup are preserved. |
| **Why not `ReconnectWithAuthKey`?** | That path is built for the *offline/stale* case: it spins up a **second** `tsnet.Server` on the same state dir and needs a disruptive `Disconnect()` first — which kills the startup-bound VPN listeners, leaving the node "connected" but unreachable for node-targeted jobs until a restart. Wrong tool for online renewal. |

The **scope is already granted**: device tokens have carried `device_authkey:write` + the `/api/fabric/authkey/generate` allowlist entry since aceteam **#4432** (citadel #383), so the node's own token can mint a fresh key — no admin path. The renewer relocates the only remaining landmine to **token vintage**: a token minted *before* #4432 will 403. We handle that gracefully (below) rather than crash a node that still self-heals on restart.

## Summary

- **`internal/network/keyhealth.go`** — pure `keyRenewalDecision(keyExpiry, now, threshold)` classifier (`KeyNoExpiry`/`KeyHealthy`/`KeyRenewSoon`/`KeyExpired`). **nil/zero `KeyExpiry` = non-expiring durable-baseline key = no-op**, not "unknown, act." Plus `InspectGlobalNodeKey` (reads `Status().Self.KeyExpiry`/`Expired`) and `RenewGlobalNodeKey` (in-place `LocalClient.Start` reauth). Default renewal threshold: 24h before expiry.
- **`cmd/keyrenew.go`** — background renewer loop started by `citadel work`, **gated like the footprint sampler**: `--no-key-renew` flag or `CITADEL_KEY_RENEW_INTERVAL<=0`. No-op in direct-Redis mode (no device token). On each tick: inspect health; if within threshold (or expired), fetch a fresh authkey with the node's own token and reauth in place. A single miss self-heals on the next tick.
- **`internal/network/reauth.go`** — classify the 403 as `ErrAuthkeyScope` (`IsAuthkeyScopeError`) so an **old device token** (pre-#4432) surfaces the exact remedy — *"re-run `citadel init` to enable online self-renewal; this node still reconnects on restart"* — instead of failing opaquely. Refreshed the stale issue-#175 comment.

**Reconnect ordering is unchanged and already correct**: `work.go` and `attemptVPNRecovery` try the no-authkey path first (3× retry) and only reach the authkey-mint path on `ErrStaleState`; `reconnect --force` (clear state + mint) stays an explicit operator choice. "Harden" here = the 403 classification + honest operator messaging, not a reorder.

## Test plan

| Group | Coverage |
|---|---|
| `keyRenewalDecision` (table-driven) | nil expiry, zero expiry, far-future healthy, just-past-threshold healthy, exactly-at-threshold renew, within-threshold renew, exactly-now expired, past expired |
| `KeyStatus.String` | all statuses + unknown |
| `InspectGlobalNodeKey` / `RenewGlobalNodeKey` guards | nil server -> nil health no error; empty authkey + not-connected rejected |
| `FetchFreshAuthkey` 403 | classified as `ErrAuthkeyScope`; 401 is **not** misclassified |
| `keyRenewIntervalFromEnv` | unset default, positive override, `0`/negative disable, non-numeric fallback |
| Renewer wiring | no-device-token no-op; `--no-key-renew` flag registered, default false |

`go test ./...` green, `go vet` + `gofmt -l` clean.

## Manual node-test plan (why draft)

Two facts can't be verified headless and gate marking this ready:

1. **Does Nexus/Headscale populate `Self.KeyExpiry`?** On a live node running `citadel work`, inspect key expiry (`citadel status` / worker log). If it is always `nil`, keys don't expire in this deployment and the renewer is **correctly inert** (`KeyNoExpiry` no-op) — confirm that's expected before relying on it.
2. **Online renewal preserves the session.** With a node whose key does have an expiry, set `CITADEL_KEY_RENEW_INTERVAL` low (e.g. `60`) and a short threshold so the loop fires, then confirm: the "Node key refreshed before expiry" log appears, the node's **IP is unchanged**, and node-targeted jobs (`terminal_exec`, `code_*`) still route to it (VPN listeners survived).
3. **Old-token 403 fallback.** On a node whose device token predates #4432, confirm renewal logs the `ErrAuthkeyScope` remedy ("re-run `citadel init`") and the node keeps running (no crash), then re-run `citadel init` and confirm renewal succeeds.
4. **Persistent-restart, no 403.** Restart a persistent (non-ephemeral, #4584) node and confirm it reconnects via the **no-authkey** path with no authkey-mint 403.

## Related

- Epic: **#4583** (self-healing fabric node identity)
- Backend sibling: **#4584** (non-ephemeral node registration keys)
- Scope grant this depends on: aceteam **#4432** / citadel **#383** (`device_authkey:write`)
- Advances **#4313** (Fabric PKI / signed sandbox capabilities) — shared identity direction

### DIFF
--- a/cmd/keyrenew.go
+++ b/cmd/keyrenew.go
@@ -1,0 +1,167 @@
+// cmd/keyrenew.go
+// Background node-key renewal loop for the self-healing fabric node identity
+// baseline (epic #4583, citadel side of backend PR #4584).
+//
+// While a node is healthy and online, this loop periodically inspects its
+// Headscale node-key expiry and, when the key is approaching expiry, refreshes
+// it IN PLACE using the node's own device token — never the offline
+// authkey-mint recovery path. Keeping the key fresh while online means the
+// session never lapses, so an ordinary restart re-establishes through the
+// existing no-authkey reconnect and the 403 authkey-mint path is never reached.
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/aceteam-ai/citadel-cli/internal/network"
+)
+
+// keyRenewCheckInterval is how often the loop inspects key-expiry health. Key
+// lifetimes are measured in days and the renewal threshold is 24h, so an hourly
+// check is far more frequent than needed — cheap insurance that a node briefly
+// offline around its threshold still renews on a subsequent tick.
+const keyRenewCheckInterval = time.Hour
+
+// keyRenewEnvInterval overrides the check interval (in seconds) and doubles as
+// the opt-out switch: a non-positive value disables the loop entirely, matching
+// the footprint sampler's gating convention (CITADEL_FOOTPRINT_INTERVAL<=0).
+const keyRenewEnvInterval = "CITADEL_KEY_RENEW_INTERVAL"
+
+// keyRenewIntervalFromEnv resolves the check interval and whether the loop is
+// enabled. Unset uses the default; a positive value overrides; a non-positive
+// value disables.
+func keyRenewIntervalFromEnv() (time.Duration, bool) {
+	raw := os.Getenv(keyRenewEnvInterval)
+	if raw == "" {
+		return keyRenewCheckInterval, true
+	}
+	var secs int
+	if _, err := fmt.Sscanf(raw, "%d", &secs); err != nil {
+		return keyRenewCheckInterval, true
+	}
+	if secs <= 0 {
+		return 0, false
+	}
+	return time.Duration(secs) * time.Second, true
+}
+
+// startNodeKeyRenewer launches the background node-key renewal loop unless it is
+// disabled via --no-key-renew or a non-positive CITADEL_KEY_RENEW_INTERVAL. It
+// no-ops when the node is not in API mode (no device token to fetch a fresh
+// authkey with) since renewal has no admin-free path without it.
+func startNodeKeyRenewer(ctx context.Context, deviceConfig *DeviceConfig) {
+	if workNoKeyRenew {
+		Debug("node-key renewer disabled (--no-key-renew)")
+		return
+	}
+	interval, enabled := keyRenewIntervalFromEnv()
+	if !enabled {
+		Debug("node-key renewer disabled (%s<=0)", keyRenewEnvInterval)
+		return
+	}
+	if deviceConfig == nil || deviceConfig.DeviceAPIToken == "" {
+		Debug("node-key renewer disabled (no device token)")
+		return
+	}
+
+	apiBaseURL := deviceConfig.APIBaseURL
+	if apiBaseURL == "" {
+		apiBaseURL = authServiceURL
+	}
+
+	go runNodeKeyRenewer(ctx, deviceConfig.DeviceAPIToken, apiBaseURL, interval)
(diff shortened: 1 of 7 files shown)


sunapi386:
Live-test result (node 1063 `ubuntu-gpu`, read-only via the Nexus API — no VPN disruption):

**Q1 — does Nexus populate a node-key expiry? → No.** `nexus_get_node 1063` reports **`Expiry: None`**. So on the current fleet the renewer's nil/zero-`KeyExpiry` branch is the operative one: `keyRenewalDecision()` returns a no-op and the loop stays correctly inert. It won't fire and can't induce churn — it's a safety net for a future where Headscale *does* set key expiries.

**Build/tests:** `go build ./...` clean, `internal/network` + `cmd` renew/reauth/scope tests green, and `citadel work --help` shows `--no-key-renew` as expected.

**Q2 — in-place reauth preserves IP + listeners:** untriggerable here (nothing expires to trigger a renewal), and I'm deliberately not forcing a disruptive reauth on the live mesh. Stays documented-not-verified until a node runs against a Headscale config that sets expiries.

Separately worth noting for the epic: this node re-registered fresh *today* (now `1063` @ 100.64.0.61, with a stale `ubuntu-gpu-1` @ 100.64.0.5 left behind) — the ephemeral-churn landmine still biting pre-deploy, which the P0 device-auth fix (aceteam #4606, now merged) addresses.

PR #461, opened 2026-07-07
fix(security): require coordinator mTLS identity for :8080 control endpoints (#5028)

sunapi386:
## Context

Phase 0 of #5028 (founder green-lit). A security audit found that the Citadel node's status/control server on **:8080** accepts requests from **any mesh IP with no cryptographic identity** and exposes an **SSH-key-injection** endpoint. Because the live Headscale mesh ACL is currently flat, any node on the mesh — including, once multi-tenant, **another org's node** — can POST an SSH key to a peer and take over the host. This PR closes the SSH-key-injection vector by requiring a **fabric-CA-signed coordinator client certificate (mTLS)**, reusing the existing fabric mTLS PKI. No new bearer token is introduced. **PR only — do not merge; requires the backend companion below to be usable, and must be rolled out backend-first.**

### The :8080 server (`internal/status/server.go`, `cmd/work.go`)
- **Transport:** plain HTTP (`ListenAndServe`), no TLS. Binds `:8080` (all interfaces) plus an explicit tsnet **VPN listener** so it is reachable over the mesh.
- **Endpoints:**
  - Read-only, unauthenticated: `/ping`, `/status`, `/health`, `/services`, and `/gateway-cert.pem` (intentionally public — it is the bootstrap-before-trust channel for the gateway's leaf cert).
  - `/api/screenshot`, `/api/actions` — desktop control, gated by an **org-scoped bearer token** (`requireAuth`). A real credential, not the hole.
  - **`/ssh/authorized-keys` — SSH key injection (host takeover), gated by `requireVPNOrAuth`.**
  - `/agent/*` introspection + control, and provisioning/workflow routes — also gated by `requireVPNOrAuth`.
- **The hole:** `requireVPNOrAuth` **short-circuits to allow ANY request whose source IP is in `100.64.0.0/10`** (the Headscale CGNAT range) with no token. Mesh origin is not an identity. On the flat mesh, a cross-org node can call `/ssh/authorized-keys` directly and write to `~/.ssh/authorized_keys`.

### Existing fabric identity model (what we verify against)
- There **is** a fabric mTLS PKI: an **EC fabric CA**. Node leaf certs are CA-signed and carry SAN URIs `aceteam:node:<node_uid>` and `aceteam:org:<org_id>` (`python-backend/utils/fabric_ca/profile.py`). Leafs are issued through the pairing CSR flow (`fabric_pairing.py`) and the public CA chain is served at **`GET https://aceteam.ai/api/fabric/ca/chain`** (+ CRL at `/api/fabric/ca/crl`). The nexus `reenroll` service (nexus#30/#31) is a keyless verifier that chains a client leaf to this CA; nginx terminates client mTLS there.
- **Gap discovered:** citadel-cli (the node) has **no fabric-cert code today** — it does not hold the fabric CA trust root, does not hold a fabric client identity, and the mutating endpoints are plain HTTP. The only cert on the node is the gateway's **self-signed server** cert (`internal/tlscert`), which is not CA-signed and is a server identity.

### Legitimate control flow (established before changing it)
- SSH deploy: Next.js `/api/fabric/ssh-keys/deploy/[nodeId]` → **python backend `fabric_relay.py:relay_ssh_keys`** → validates org ownership in the DB → **POSTs plain HTTP over the VPN mesh** to `http://{vpn_ip}:8080/ssh/authorized-keys`. **It presents no credential.** At the node, this legitimate call is byte-for-byte indistinguishable from a cross-org attacker's — which is exactly the vulnerability.

**Consequence:** no citadel-only change can *both* fail-closed *and* preserve today's flow, because today's flow carries no credential. This PR fails closed; the backend must be updated to present a coordinator cert (see gaps).

## Phase 2 — What this PR gates

A dedicated **mTLS control listener** (default **:8443**) now serves the SSH-key-injection endpoint, gated by a verified **coordinator** client certificate:

- **`internal/status/mtls.go` — `FabricCAVerifier`:** loads the public fabric CA bundle from a configured path, exposes a `RequireAndVerifyClientCert` TLS config (chain-to-CA enforced at the handshake), and a `requireCoordinator` middleware that additionally checks the leaf's SAN URI is on an **explicit coordinator allowlist**.
- **`internal/status/server.go`:** the plain :8080 route for `/ssh/authorized-keys` is replaced with a **fail-closed 403 stub** (`handleSSHMovedToControl`, no side effects); the real handler is registered only on the mTLS control mux. Adds `AddControlListener` (VPN) + `startControlServer`.
- **`cmd/work.go`:** builds the verifier + a control server cert and starts the listener (local + VPN) **only when** `CITADEL_FABRIC_CA_BUNDLE` and `CITADEL_COORDINATOR_SANS` are set; otherwise logs and leaves the mutating endpoints refused.

### Fail-closed behavior (the load-bearing property)
Every one of these refuses with **zero side effects** (no key written):
- No fabric CA bundle / empty CA / **empty coordinator allowlist** → verifier refuses to construct → control listener never starts → SSH endpoint refused everywhere.
- Caller with **no client cert** → rejected at the TLS handshake.
- Client cert that **does not chain** to the fabric CA (e.g. another CA) → rejected at the handshake.
- Valid fabric cert whose SAN is **not** an allowlisted coordinator (e.g. a peer org's node cert) → **403**.
- Old plaintext/VPN-origin path → **403** stub.

### Deliberate deviation from the task's suggested interim
The task allowed gating to "any valid fabric cert" if coordinator can't be distinguished. **That does not close #5028:** the threat is other orgs' nodes, and every enrolled node holds a valid fabric leaf, so "chains to the CA" would still admit them. This PR therefore **requires an explicitly-configured coordinator identity and refuses to construct a verifier without one** (fail closed) rather than accept any fabric cert.

### Scope choice
Gated **SSH-key injection** (the host-takeover crown jewel, smallest-blast-radius migration — one caller). `/agent/*` control + provisioning/workflow share the same `requireVPNOrAuth` weakness and should move to the same gate in an immediate fast-follow; they are higher-blast-radius (many backend flows) and warrant their own coordinated rollout. Read-only endpoints stay plain (recommended: leave unauthenticated; they leak no control). Desktop endpoints keep their org-token gate.

## Test plan
`go build ./...`, `go vet ./...` clean; `go test ./internal/status/... ./cmd/...` green.
- `mtls_test.go` (synthetic EC CA fixture): verifier fail-closed construction (empty path / missing file / certless bundle / empty allowlist); `requireCoordinator` handler fail-closed (no TLS, no peer cert, peer-org SAN → 403, coordinator SAN → pass); **end-to-end over a real mTLS handshake on `/ssh/authorized-keys`** — no cert → handshake reject (no write); untrusted-CA cert → handshake reject (no write); valid fabric cert, non-coordinator SAN → 403 (no write); coordinator cert → 200 + key written.
- `ssh_handler_test.go`: rewrote the endpoint test to assert the plaintext path now fail-closes (VPN origin **and** valid org token → 403, nothing written); kept request-validation coverage.

## Flagged gaps / cross-repo dependencies (required for this fix to function)
1. **Backend companion (aceteam):** `fabric_relay.py:relay_ssh_keys` must present a coordinator fabric client cert when POSTing to the node's control listener (currently plain HTTP, no cert). **Until then, SSH deploy fails closed.**
2. **CA profile gap (aceteam):** `utils/fabric_ca/profile.py` has only `aceteam:node:`/`aceteam:org:` SANs — **no coordinator/role SAN**. The coordinator identity must be configured out-of-band (`CITADEL_COORDINATOR_SANS`) until the CA mints a distinct coordinator role.
3. **Node CA-bundle provisioning:** the node must obtain the public fabric CA bundle (`GET /api/fabric/ca/chain`) into `CITADEL_FABRIC_CA_BUNDLE`. Not yet wired; the verifier reads it once at startup (no hot-reload on CA rotation — restart required).
4. **Rollout order:** ship (1) backend-first, then update nodes — otherwise SSH-deploy fails closed fleet-wide. Node server-cert trust for the coordinator's outbound call is also unresolved (the control cert is self-signed today).

### Founder decision surfaced
This is the **node-side half** of a two-part mTLS fix, consistent with "use the existing fabric mTLS identity, no new bearer token." If a citadel-only stopgap deployable **without** the backend change is wanted, the only option is narrowing `requireVPNOrAuth` from all of `100.64.0.0/10` to an **allowlist of the coordinator/relay mesh IP** — fail-closed against cross-org, no cert needed, but it diverges from the mTLS directive and is brittle to mesh-IP churn. Offered as your call, not the default.

---

### Adjacent surfaces reviewed (confirmed not additional vectors)
- `internal/nexus/sshkeys.go` — the **outbound** key sync: the node pulls its authorized_keys from the AceTeam API via a node-authenticated Bearer `GET` and writes them locally. No inbound listener; a peer cannot trigger it. Different (node→API bearer) trust model, not the mesh hole.
- `internal/network/server.go` — the tsnet transport wrapper (Connect / IP / Disconnect). No host-mutating HTTP endpoint exposed to peers.
- `internal/network/keyhealth.go` — WireGuard/tsnet **node-key** expiry health + in-place renewal using the node's own session (epic #4583). Unrelated to authorized_keys or inbound requests.

So the SSH-injection surface reachable from a peer is the single `/ssh/authorized-keys` HTTP endpoint gated here. The remaining `requireVPNOrAuth` mutating endpoints (`/agent/set-log-level`, `/agent/worker-restart`, `/agent/resubscribe`, provisioning/workflow) stay VPN-origin-gated — cross-org-reachable but not host-takeover — pending the fast-follow; **:8080 is not "fully secured" by this PR**, only the SSH-key-injection takeover vector is closed.

### DIFF
--- a/cmd/work.go
+++ b/cmd/work.go
@@ -14,6 +14,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -1149,6 +1150,48 @@ func runWork(cmd *cobra.Command, args []string) {
 			}
 		}
 
+		// Gate the MUTATING control endpoints (SSH-key injection) behind a
+		// coordinator mTLS client certificate on a dedicated control listener
+		// (issue #5028). Enabled only when a fabric CA bundle AND coordinator
+		// identities are configured. If construction fails, we log and leave the
+		// control listener OFF -- the mutating endpoints then stay refused on the
+		// plaintext path (fail closed), never reverting to VPN-origin trust.
+		controlPortForVPN := 0
+		if caBundle := strings.TrimSpace(os.Getenv("CITADEL_FABRIC_CA_BUNDLE")); caBundle != "" {
+			coordinatorSANs := splitAndTrim(os.Getenv("CITADEL_COORDINATOR_SANS"))
+			verifier, verr := status.NewFabricCAVerifier(caBundle, coordinatorSANs)
+			if verr != nil {
+				fmt.Fprintf(os.Stderr, "   - ⚠️ Coordinator mTLS control listener NOT enabled (%v); SSH-key injection stays refused\n", verr)
+			} else {
+				var ctrlIPs []net.IP
+				if ip4, _ := network.GetGlobalIPv4(); ip4 != "" {
+					if ip := net.ParseIP(ip4); ip != nil {
+						ctrlIPs = append(ctrlIPs, ip)
+					}
+				}
+				ctrlCert, certErr := tlscert.EnsureCert(tlscert.Config{
+					Hostname:    nodeName,
+					IPAddresses: ctrlIPs,
+					CertDir:     filepath.Join(platform.ConfigDir(), "control-tls"),
+				})
+				if certErr != nil {
+					fmt.Fprintf(os.Stderr, "   - ⚠️ Coordinator mTLS control listener NOT enabled (server cert: %v)\n", certErr)
+				} else {
+					controlPort := status.DefaultControlPort
+					if p := strings.TrimSpace(os.Getenv("CITADEL_CONTROL_PORT")); p != "" {
+						if parsed, perr := strconv.Atoi(p); perr == nil && parsed > 0 {
+							controlPort = parsed
+						}
+					}
+					serverCfg.CAVerifier = verifier
+					serverCfg.ControlServerCert = &ctrlCert
+					serverCfg.ControlPort = controlPort
+					controlPortForVPN = controlPort
+					fmt.Printf("   - Coordinator mTLS control listener enabled on :%d for mutating endpoints (#5028)\n", controlPort)
+				}
+			}
+		}
+
 		statusServer := status.NewServer(serverCfg, collector)
 
 		// Register provisioning API routes on the status server
@@ -1174,6 +1217,20 @@ func runWork(cmd *cobra.Command, args []string) {
 				statusServer.AddListener(vpnLn)
 				Log("status server VPN listener on %s:%s", vpnIP, vpnPort)
 				fmt.Printf("   - Status server VPN: http://%s:%d\n", vpnIP, workStatusPort)
+			}
+
+			// Also expose the mTLS control listener over the mesh so the
+			// coordinator can reach the mutating endpoints via the VPN (#5028).
+			if controlPortForVPN > 0 {
+				ctrlPort := fmt.Sprintf("%d", controlPortForVPN)
+				if ctrlLn, ctrlIP, err := network.ListenVPN("tcp", ctrlPort); err != nil {
+					Log("control server VPN listener failed (LAN-only): %v", err)
+					fmt.Fprintf(os.Stderr, "   - ⚠️ Control server VPN listener failed (LAN-only): %v\n", err)
+				} else {
+					statusServer.AddControlListener(ctrlLn)
+					Log("control server VPN listener on %s:%s", ctrlIP, ctrlPort)
+					fmt.Printf("   - Control server VPN (mTLS): https://%s:%d\n", ctrlIP, controlPortForVPN)
+				}
 			}
 		}
 
@@ -2374,4 +2431,17 @@ func initProvisionHandler() *provision.Handler {
 	// was down and updates their status accordingly.
 	mgr.ReconcileAll(context.Background())
 	return provision.NewHandler(mgr)
+}
+
+// splitAndTrim splits a comma-separated env value into non-empty, trimmed items.
+// Used for CITADEL_COORDINATOR_SANS (the allowlist of coordinator SAN URIs that
+// may invoke mutating control endpoints -- issue #5028).
+func splitAndTrim(s string) []string {
+	var out []string
+	for _, part := range strings.Split(s, ",") {
+		if p := strings.TrimSpace(part); p != "" {
+			out = append(out, p)
+		}
+	}
+	return out
 }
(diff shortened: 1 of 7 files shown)


sunapi386:
### Added: node CA-bundle provisioning + default coordinator SAN (part 2 of #5028)

Follow-on commit on this branch so the mTLS control listener is actually usable end-to-end (paired with backend PR aceteam-ai/aceteam#5069).

- **`internal/status/ca_bundle.go` — `EnsureFabricCABundle(baseURL, configDir)`**: fetches the public fabric CA chain from `GET /api/fabric/ca/chain` at startup and caches it to `ConfigDir()/fabric-ca-bundle.pem`, so the listener has a trust root without manual `CITADEL_FABRIC_CA_BUNDLE` config. Read once at startup; **a CA rotation requires a node restart to re-fetch.**
- **Zero-break landmine fix**: because this branch turns plaintext `:8080/ssh/authorized-keys` into a 403 stub, a transient backend outage at startup would otherwise disable SSH-deploy entirely (no :8443, and :8080 is a stub). So a fetch failure **reuses a previously-cached bundle** rather than gating listener bring-up on a live fetch. A non-PEM/error response never clobbers a good cache. Only a first-ever cold start with no cache and an unreachable backend leaves the listener down (fail closed, documented).
- **`DefaultCoordinatorSAN = "aceteam:coordinator"`**: `cmd/work.go` now defaults `CITADEL_COORDINATOR_SANS` to this when unset, so upgraded nodes allowlist the platform coordinator out of the box. Without at least one SAN the verifier refuses to construct and the listener stays off.

**Cross-repo SAN contract**: `DefaultCoordinatorSAN` MUST equal the backend's `profile.COORDINATOR_URI` (`"aceteam:coordinator"`). Pinned byte-for-byte on both sides — `internal/status/mtls_test.go::TestDefaultCoordinatorSANContract` here, `test_fabric_ca_authority.py::TestIssueCoordinatorLeaf` there. (Note: the pre-existing `coordinatorSAN` test fixture in `mtls_test.go` uses `aceteam:role:coordinator` purely as a generic example of the env-driven mechanism; the real default is `aceteam:coordinator`.)

Tests: `internal/status/ca_bundle_test.go` (fetch+cache, cache reuse on failure, cold-start error, bad-response-no-clobber, non-200). `go build/vet/test ./internal/status/` green.

---

PR #487, opened 2026-07-10
feat(worker): MODULE_SET handler + live ModuleOps adapter (interim, Option A)

sunapi386:
## Context

The `node_module_set` MCP tool (aceteam-ai/aceteam#5280) already lets an operator
assign a single module a desired state on a specific node. It xadds a `MODULE_SET`
job to the node's per-node stream carrying one `ModuleAssignment`
(`{source, desired_status: running|stopped|absent, config}`), but **no citadel
handler consumed it** — the job nacked and dead-lettered. This PR closes that gap
with an interim, imperative handler that reuses the existing (tested-but-dormant)
`internal/reconcile` engine.

The full design (#4273) is a declarative pull loop where the node fetches its
whole desired set from a control-plane serve endpoint that does not exist yet.
This is **Option A**: apply the one assignment now, converging into #4273 later.

## What it does

- **`internal/worker/module_set.go`** — a privileged, per-node `MODULE_SET`
  handler. It fails closed unless the job arrives on the per-node stream
  (`isPerNodeStream`), parses the single `ModuleAssignment` from the flattened
  payload, runs the scoped reconcile + apply, and returns success / retry.
- **`cmd/module_ops.go`** — the live `reconcile.ModuleOps` adapter (the bulk of
  the work), mapping install/uninstall/start/stop/list onto the existing
  catalog + compose + lockfile + manifest machinery.

### The single-module-scoped-actual trick

The reconcile engine treats `desired` as authoritative for the **whole node** —
anything in `actual` but not in `desired` is uninstalled. Passing the node's full
installed set as `actual` with a one-entry `desired` would uninstall every other
module. So the handler scopes **both** sides to just the target module: it lists
installed modules, finds the one whose canonical `Source` equals the assignment's
`Source`, and passes only that as `actual` (empty when not installed). `absent` is
realized as an **empty desired set** for the scoped module, which the engine
converges by uninstalling it (a no-op if not installed). The engine then emits
exactly the one step for this module and touches nothing else.

### The net-new uninstall primitive

No imperative uninstall existed before. `liveModuleOps.Uninstall` is per-module and
idempotent: `docker compose down` → drop the service from `citadel.yaml` →
`catalog.DeleteLockEntry` (new) → remove the materialized compose/sandbox/env
files. Uninstalling a module that is not installed is a no-op success. A real
`compose down` failure (docker daemon down) returns a transient error so the job
retries rather than de-registering a still-running stack.

### How `stopped` is made durable

`cmd/run.go` (`runAllServices`) and `cmd/work.go` (`startManagedServices`) compose
up every manifest service on boot, so a naive imperative stop comes back on the
next `citadel work` restart. This adds a per-service `desired_status` marker to the
manifest `Service`; `Stop` sets it to `stopped` (marker first, then compose down),
and both boot paths **skip** services where `serviceStartDisabled(s)` is true. The
module stays installed but is not started — `stopped` is distinct from `absent`,
and it survives a reboot. `Start`/`Install` clear the marker.

### The `name`-gap note (aceteam-side fix warranted)

The payload has **no `name`**, so the reconcile key falls back to the source
basename (`NameFromSource`), which need not equal the module's declared service
name in `citadel.yaml`. Keying naively on the basename would churn (re-install
every pass) and orphan the real service. The handler avoids this by **matching on
canonical `Source`** and aligning the desired assignment's `Name` to the installed
service's real name before reconciling, so start/stop/uninstall carry a name the
adapter maps back to a manifest service. `ListInstalled` reports `Source` in the
requested canonical form (lockfile `src.Raw`), and `LockEntry.Config` (new) is
recorded/reported so a converged node does not see spurious config drift.

This is a workaround. `internal/reconcile/types.go` itself says the control plane
**SHOULD** send `Name` for stable keying. An aceteam-side fix — having
`node_module_set` include the resolved module name in the assignment — is
**warranted** and would remove the one remaining fragile case (a source-ref change
where the repo's `service.yaml` name also changes across refs → orphan). Not fixed
here (this PR does not touch the aceteam repo).

## Test plan

Scoped tests only (never `go test ./...`, which crashes this env):

| Area | Coverage |
|------|----------|
| `internal/worker` (`TestModuleSet*`, 9 tests) | fail-closed on shared pool; install missing; **does not touch other installed modules**; absent uninstalls + idempotent; **stopped stops (not uninstall) and stays installed**; converged running = no-op (no churn, incl. name != basename); missing source terminal; list/install failure → retry |
| `internal/reconcile/...` | existing suite still green |
| Build | `go build ./cmd ./internal/worker ./internal/reconcile ./internal/catalog` + full `./cmd/citadel` binary links |

The live adapter (`cmd/module_ops.go`) is build-verified: its container-touching
ops are injectable seams, but unit-testing it directly would read/write the host's
real config dir, and running `./cmd` tests risks the crash-prone tree. Its core
routing/scoping/idempotency/absent/stopped semantics are fully exercised through
the handler tests against a stateful fake `reconcile.ModuleOps`. The `run.go` /
`work.go` boot-skip is build-verified and driven by the single shared predicate
`serviceStartDisabled`.

## Related

- Interim per aceteam-ai/aceteam#5280; converges into #4273 durable desired-state.
- Mirrors the AGENT_UPDATE (#4427) / WHATSAPP_PROVISION (#4454) per-node privileged
  handler pattern.

---

### DIFF
--- a/cmd/manifest.go
+++ b/cmd/manifest.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/aceteam-ai/citadel-cli/internal/platform"
 	"github.com/aceteam-ai/citadel-cli/services"
@@ -17,6 +18,22 @@ type Service struct {
 	Type        string `yaml:"type,omitempty"`         // "native" or "docker" (default: auto-detect)
 	ComposeFile string `yaml:"compose_file,omitempty"` // For docker services
 	Port        int    `yaml:"port,omitempty"`         // For native services
+	// DesiredStatus is the operator-assigned run-state for boot. Empty means the
+	// service is started on boot (the historical behavior). "stopped" makes a
+	// remote MODULE_SET "stopped" DURABLE: the service stays installed but the
+	// boot paths (citadel run / citadel work) SKIP composing it up, so a stop
+	// survives a reboot instead of silently coming back. This is a per-service
+	// marker; it is NOT the same as uninstalling (which removes the service).
+	DesiredStatus string `yaml:"desired_status,omitempty"`
+}
+
+// serviceStartDisabled reports whether a service is marked "stopped" and must be
+// SKIPPED by the boot-time service-start paths (runAllServices in cmd/run.go and
+// startManagedServices in cmd/work.go). This is the single predicate both boot
+// paths consult so a remote-assigned "stopped" state is honored consistently and
+// does not restart on reboot.
+func serviceStartDisabled(s Service) bool {
+	return strings.EqualFold(strings.TrimSpace(s.DesiredStatus), "stopped")
 }
 
 // ManifestCapabilities defines the optional capabilities section in citadel.yaml.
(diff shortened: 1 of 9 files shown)


sunapi386:
Follow-up commit d21a089 fixes an ordering landmine in the live adapter: `Install` previously ran `Uninstall` (compose down + de-register + delete lockfile + rm files) **before** the network-dependent `resolveModuleForTUI`/`VerifyModule`. On an update (config drift, or re-assert), a transient clone/verify failure would delete the running module and, on retry, leave it uninstalled. Now resolve+verify run while the old module is still up, and update-detection keys on the resolved `manifest.Name` (also fixes source-ref changes where the service name differs from the basename).

Residual caveat (interim, acceptable): if that in-place `Uninstall` succeeds but the subsequent `InstallFromManifest` fails, the module is left down until the job retries and reinstalls it. Full transactional update is out of scope for the interim handler and converges into #4273.

---

Issue #495, opened 2026-07-11
TUI Chat tab: connect to AceTeam Team Chat (parity with web/iOS/Android)


sunapi386:
## Context

The Control Center's Chat tab is currently a **node-to-node toy chat**, not AceTeam Team Chat. `internal/tui/controlcenter/chat_page.go` + `internal/chat/client.go` implement an ephemeral org-scoped pub/sub chat over the Redis API proxy (`chat:v1:org_{org}:general` via `/api/fabric/redis/*`): one hardcoded `#general` channel, presence heartbeats, **no history** (messages are written to a Redis stream nobody reads back), and no connection to the product's Team Chat that users see on web (`/team-chat`), iOS, and Android.

Meanwhile Team Chat is a real, cross-platform product surface backed by Supabase (`channels`, `channel_messages`, `channel_members`) with agents, threads, attachments, unread state, and MCP tools. The TUI should mirror it: a user standing at a terminal on their node should see and participate in the same org conversation as everyone else.

## Target UX (parity with web/iOS/Android)

- **Channel list**: org channels + DMs the node's user can access, with unread indicators
- **Read messages**: history on open (paginated), live updates while viewing
- **Send messages**: post as the authenticated user; failures surfaced inline
- **Member list**: members of the active channel
- **Search**: text search within a channel
- Keep the existing node-to-node chat available (it has a distinct purpose: node presence/ops chatter that works with only a device token)

## Backend contract (verified in aceteam repo)

Both mobile clients use the **v1 REST surface** under `app/api/channels/**` with `Authorization: Bearer <token>`; models mirror `types/channels.ts`. iOS (`ios/AceTeam/Services/TeamChatService.swift`) and Android (`android/.../data/api/AceTeamApi.kt`) do **light polling (~5s)** for live updates — the v1 web WebSocket (`/api/channels/ws`) has been deleted and its server-side broadcast is a no-op pending the v2 cutover, and the live v2 socket (`/api/chat/ws`) serves the separate `chat_*` conversations surface. So **polling is the current parity mechanism**, matching mobile.

| Method | Path | Request | Response |
| --- | --- | --- | --- |
| GET | `/api/channels` | — | `{ channels: Channel[] }` (auto-seeds `#general`) |
| GET | `/api/channels/{id}/messages` | `?cursor&limit(1-100,def 50)&direction(before\|after)` | `{ messages: ChannelMessageWithSender[], hasMore, nextCursor? }` |
[...]

PR #498, opened 2026-07-11
feat(tui): Team Chat tab — real AceTeam org channels in the Control Center

sunapi386:
## Context

The Control Center's "Chat" tab has been a node-to-node toy: ephemeral pub/sub over the Redis proxy, one hardcoded `#general`, no history, zero connection to the Team Chat users actually live in on web/iOS/Android. This PR makes the TUI a real Team Chat client — the same org channels, messages, members, and agents everyone else sees — rendered in the terminal.

Contract verified against the aceteam repo before writing any code: both mobile apps consume the v1 REST surface (`/api/channels/**`, shapes in `types/channels.ts`) with a Bearer token and do ~5s light polling for live updates (the v1 chat WebSocket is deleted server-side and the v2 socket serves the separate `chat_*` conversations model — polling **is** the current cross-client parity mechanism). Full investigation in #495.

## Summary

**`internal/teamchat` — typed REST client** (commit 1)
- `ListChannels`, `Messages` (cursor paging), `SendMessage` (incl. threaded replies), `ListMembers`, `SearchMessages`, `MarkRead`, `UnreadCounts` against `/api/channels/**`
- Tolerant `Timestamp` parsing (Supabase `+00:00` offsets, fractional seconds, offset-less)
- `*APIError` preserves HTTP status; `IsAuthError` distinguishes 401/403 so the UI can render setup guidance instead of a raw scope-denial — this matters because a node's **device token is endpoint-whitelisted away from `/api/channels/**`** (403 SCOPE_DENIED) until the #495 backend follow-up lands
- `ResolveToken` mirrors the `citadel mcp` credential chain: `ACETEAM_API_KEY` env > new `aceteam_api_key` config.yaml field > device token

**Team Chat TUI page** (commit 2)
- Channel list (keyboard + mouse selection), message history, send-as-user, member sidebar (admins highlighted), `/search <text>` command, automatic mark-read on newest message
- 5s polling while the tab is active, stopped on deactivate (mobile parity; ID-based state means no echo-dedup hacks like the pub/sub page needs)
- Same keyboard contract as the existing chat page (reuses `classifyChatKey` / `nextChatPaneIndex`): Tab cycles panes and bubbles at the edges, Escape releases the input, PgUp/PgDn scroll
- On auth failure the page renders actionable setup steps (generate key at Settings → API Keys, set `ACETEAM_API_KEY` or `aceteam_api_key`), rendered once rather than per poll tick (commit 3)
- Existing node-to-node tab preserved, retitled **"Node Chat"** (registered name `chat` unchanged, so `ShowChat()` and settings wiring are untouched)

## What works end-to-end vs. gated

| Path | Status |
| --- | --- |
| User `act_` API key (env or config.yaml) | Works today — default keys have `scopes: ["*"]`, `allowedEndpoints: ["*"]` |
| Node's device token | Blocked by design: 403 from the device-key endpoint whitelist → page shows setup guidance. Backend follow-up (extend device-key `allowedEndpoints` + `team_chat` scopes) tracked in #495 Phase 2 |

## Test plan

| Group | Coverage |
| --- | --- |
| `internal/teamchat` client (12 tests) | every endpoint against `httptest` fakes: auth header, query/body encoding, pagination cursors, parent-message omission, 401/403→`IsAuthError`, 500 is not auth, timestamp format matrix, sender-label fallbacks |
| `internal/teamchat` token (4 cases) | credential chain priority: env > config > device > none |
| TUI page pure functions | message formatting (human/agent colors, thread marker, reply count, attachments, tview-markup escaping), `/search` parsing, newest-ID selection, Page interface identity, Node Chat retitle |
| Manual (needs a real key) | run `ACETEAM_API_KEY=act_... citadel` on an enrolled node → Team Chat tab: channels load, history renders, send appears on web `/team-chat`, poll picks up messages sent from web, guidance page on a device-token-only node |

`go build ./...` and the full `go test ./...` suite are green.

## Related

- Closes nothing yet — initial implementation for #495 (Phase 1 of 3)
- #495 Phase 2: aceteam backend — allow device tokens onto `/api/channels/**` so Team Chat works with zero configuration
- #495 Phase 3: fold Node Chat into the Team Chat page as a local channel; adopt the v2 WebSocket when the platform's chat migration lands

### DIFF
--- a/cmd/controlcenter.go
+++ b/cmd/controlcenter.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aceteam-ai/citadel-cli/internal/platform"
 	pmx "github.com/aceteam-ai/citadel-cli/internal/proxmox"
 	"github.com/aceteam-ai/citadel-cli/internal/status"
+	"github.com/aceteam-ai/citadel-cli/internal/teamchat"
 	"github.com/aceteam-ai/citadel-cli/internal/telemetry"
 	"github.com/aceteam-ai/citadel-cli/internal/terminal"
 	"github.com/aceteam-ai/citadel-cli/internal/tui"
@@ -243,7 +244,10 @@ func runControlCenter() {
 		// longer shows a permanent "device authorization required" status.
 		Chat:               buildChatConfig(),
 		ChatConfigProvider: buildChatConfig,
-		Proxmox:            buildProxmoxConfig(),
+
+		TeamChat:               buildTeamChatConfig(),
+		TeamChatConfigProvider: buildTeamChatConfig,
+		Proxmox:                buildProxmoxConfig(),
 		Settings: controlcenter.SettingsCallbacks{
 			LoadTelemetry: func() *config.Telemetry {
 				return config.LoadTelemetry(platform.ConfigDir())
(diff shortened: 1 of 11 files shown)


sunapi386:
Live-verified the team-chat client against production (`https://aceteam.ai`) with a user `act_` API key. All endpoints return 200 and decode cleanly into the client structs:

- `GET /api/channels` → 10 channels
- `GET /api/channels/{id}/messages` → decoded (page + hasMore)
- `GET /api/channels/{id}/members` → hydrated user join works (admin/member roles resolved to names)
- `GET /api/channels/unread-counts` → 10 channels tracked
- `GET /api/channels/{id}/messages/search` → OK

Auth works with a user `act_` key (via `ACETEAM_API_KEY` env or the `aceteam_api_key` config field). The device-token path remains gated on #495 (Phase 2 backend: extend device-key `allowedEndpoints` + `team_chat` scopes) and shows setup guidance until then. Moving out of draft.

Issue #501, opened 2026-07-11
work mode has no desktop path on the mesh: deskstream (5910) never starts, VNC (5900) has no mesh listener — controlcenter-only today


sunapi386:
## Context

Fleet-wide web-desktop outage diagnosed in aceteam-ai/aceteam#5419: the platform relay dials raw TCP `mesh_ip:5900` (VNC) and `mesh_ip:5910` (H.264 deskstream), but `citadel work` — what every fleet node runs — binds tsnet mesh listeners only for 7860/8080/control/8443. The embedded VNC server + deskstream server + their `ListenVPN` mesh listeners exist only in `citadel controlcenter` (TUI) mode. So headless nodes have never served web desktop.

## Asks

1. **H.264 deskstream in work mode**: start the deskstream server (or expose it via a gateway route) when the `desktop` permission is enabled, so capability-flagged H.264 nodes (`capabilityFlags.h264`) actually serve it headless.
2. **Decide the canonical desktop transport** and align both repos: either (a) the gateway `/vnc` route on 8443 (already works in work mode; the platform relay is being re-pointed at it per #5419) — then the raw 5900/5910 mesh dials die; or (b) raw mesh listeners for 5900/5910 in work mode (parity with controlcenter). (a) is preferred: TLS + permission-gated + no new listeners.
3. **macOS**: `EnsureGatewayVNC` is Linux-only (x11vnc). On macOS nothing listens on 5900 unless the user enables Screen Sharing — the websockify bridge accepts and then fails per-connection. Either document/surface "enable Screen Sharing" as a first-class node-side status (so the web UI can show an actionable error), or ship an embedded VNC provider for macOS work mode.

## Evidence

- `cmd/work.go`: `ListenVPN` sites = status, control, terminal (7860), gateway (8443) only; websockify bridge binds `127.0.0.1:6080`; `EnsureGatewayVNC` no-op off Linux.
- `cmd/controlcenter.go`: `ccVNCPort=5900` + `ListenVPN`, `ccH264Server` (deskstream, DefaultPort 5910) — TUI mode only.
- Live on jmbp16 (`citadel work`, v2.71.x): 6080 loopback + 8443 up, nothing on 5900/5910, relay close 4013 both ports.

---
*Generated by Claude Fable 5*

sunapi386:
[...]

PR #505, opened 2026-07-11
fix(work): move mTLS control listener off mesh :8443; fail loud on gateway mesh-bind loss (#504)

### LINKED ISSUE #504: work mode: mTLS control listener and HTTPS gateway both default to mesh :8443 — gateway loses the bind, mesh /vnc and /modules/* unreachable
## Symptom (live on the fleet, 2026-07-11)

`citadel_logs` on BOTH jmbp16 (node 1083) and ubuntu-gpu (node 1084):

```
# jmbp16, 100.64.0.21
gateway VPN listener failed (LAN-only): listen on 100.64.0.21:8443: tsnet: listener already open for tcp, 100.64.0.21:8443
# ubuntu-gpu, 100.64.0.24
gateway VPN listener failed (LAN-only): listen on 100.64.0.24:8443: tsnet: listener already open for tcp, 100.64.0.24:8443
```

## Root cause

`cmd/work.go` stands up two independent listeners that both default to mesh port **8443**:

1. The **mTLS control listener** (`internal/status/server.go`, `DefaultControlPort=8443`, `RequireAndVerifyClientCert`) — added in #5028's platform work (commit `d8eaa8f`), binds the mesh IP **first**.
2. The **HTTPS gateway** (`internal/gateway/gateway.go`, default 8443; serves `/vnc`, `/terminal`, `/modules/*`) — its `ListenVPN` runs later and **loses the bind**, logging the warning above and silently degrading to LAN-only.

Net effect: over the mesh, `:8443` answers with an mTLS server that hard-rejects certless clients at the TLS handshake. So:

- **`/vnc` over the mesh is unreachable** — this gates the platform-side desktop fix (aceteam-ai/aceteam#5419); the planned relay→gateway re-point cannot ship until this is resolved.
- **`/modules/*` over the mesh is likely broken again** — the WhatsApp bridge mesh calls (`whatsapp_health`, `whatsapp_list_contacts`, fixed in aceteam#4709 to dial `https://mesh:8443/modules/...`) would now hit the mTLS listener and fail TLS. Worth re-testing on prod.

Related: #501 (work-mode desktop parity) — this collision is the immediate gate in front of option (a) there.

## Fix options

1. **Separate default ports** (simplest): move the control listener off 8443 (e.g. 8444) — coordinator-side dial config must follow (aceteam repo), so coordinate the rollout.
2. **Fold the control routes into the gateway** with per-route TLS/client-cert requirements, so one listener owns 8443.
3. At minimum: **fail loud** — the silent `LAN-only` fallback hid this on every node since the control listener shipped; a port collision between two of our own listeners should be a startup error, not a warning.

## Latent-landmine note

Classic silent-degradation: both listeners "work" in isolation and the node reports healthy; only a mesh-side dial of a gateway route exposes the loss. Startup should assert that the gateway's mesh listener actually owns its port.

---
*Generated by Claude Fable 5*

sunapi386:
## Context

Fixes #504. Every node in the fleet is currently serving the wrong thing on mesh `:8443`: `cmd/work.go` starts the mTLS control listener (`internal/status`, added for SSH-key injection in aceteam#5028) and the HTTPS gateway (`internal/gateway`, serving `/vnc`, `/terminal`, `/modules/*`) with the **same default mesh port 8443**. The control listener binds the mesh IP first, the gateway's tsnet bind fails with `tsnet: listener already open`, and the gateway silently degrades to LAN-only while the node keeps heartbeating healthy.

Confirmed blast radius (see #504): the desktop relay re-point (aceteam#5419) is gated, and WhatsApp bridge mesh calls (`whatsapp_health` etc., which dial `https://mesh:8443/modules/whatsapp/...` per aceteam#4709) fail TLS against the mTLS listener — "certificate still failed verification after refreshing it" — fleet-wide. Related: #501 (work-mode desktop parity, option (a) is gated behind this).

This PR implements options **1 + 3** from the issue: separate the default ports, and make the gateway's mesh-bind loss unmissable.

## Summary

**1. Control listener moves to mesh `:8444`, registered in the port registry** (`services/ports.go`)

- New `services.ControlMTLSPort = 8444`, added to `ReservedCitadelPorts` (`"control-mtls"`), so the existing apps/module host-port collision guard (`internal/apps/hostport_collision_test.go`) now covers it and no module/app can ever be allocated it.
- `internal/status.DefaultControlPort` now *derives from the registry* instead of hardcoding a literal — the registry is the single source of truth for citadel-owned ports, which is exactly the discipline whose absence caused this collision. `CITADEL_CONTROL_PORT` env override is unchanged.
- Gateway stays on `:8443` — that URL shape is baked into the coordinator's module/VNC/terminal dial paths and the gateway-cert trust flow; moving the *internal-only* control listener is the far smaller blast radius.

**2. Loud failure on gateway mesh-bind loss** (`cmd/work.go`, `cmd/gateway_bind_error.go`)

- The gateway's `ListenVPN` failure escalates from a one-line `LAN-only` warning to a multi-line startup **error** on stderr + log: it names the port, spells out what is now unreachable (`/vnc`, `/terminal`, `/modules/*` over the mesh; desktop/WhatsApp/modules platform features), and — when the error is tsnet's same-process `listener already open` — names the culprit knobs (`--gateway-port` vs `CITADEL_CONTROL_PORT`, default 8444). A collision between two of our own listeners must never be silent again.
- The process **keeps booting**: LAN service is intact, and killing the node over a mesh bind would be a worse failure mode. The escalation is about visibility, not fatality.

## Rollout compatibility

The coordinator (aceteam python-backend, `routes/fabric_relay.py`) dials node control at mesh `:8443` today. Companion PR aceteam-ai/aceteam#5450 changes the dial to **try `:8444`, then legacy `:8443`, then the existing flag-gated plaintext `:8080`**. A heartbeat-advertised control port was considered and rejected: it needs node → heartbeat → node-status-worker → persistence → relay plumbing (including a DB column) for a value with exactly two possible states, while try-with-fallback is self-healing across every skew case.

Version-skew matrix (SSH-key relay, the only control-port consumer):

| Coordinator | Node | `:8444` dial | `:8443` dial | Outcome |
| --- | --- | --- | --- | --- |
| old (8443 only) | old (control on 8443) | — | control listener answers | **works** (status quo) |
| old (8443 only) | **new** (control on 8444, gateway on 8443) | — | hits the **gateway**; its self-signed cert fails fabric-CA verification → transport error → plaintext `:8080` fallback → 403 stub | **fails closed** with a clear 403 until the coordinator updates → deploy aceteam#5450 **before** releasing this |
| **new** (8444→8443) | old (control on 8443) | connection refused (nothing there) → next | control listener answers | **works** |
| **new** (8444→8443) | **new** | control listener answers | not attempted | **works** |
| **new** (8444→8443) | pre-#5028 (no control listener at all) | refused | refused / TLS error | plaintext `:8080` fallback (flag-gated, unchanged) — **works** |

Key safety property: a control dial can never get a *false positive* from the gateway — the coordinator's mTLS client verifies the server cert against the fabric CA, and the gateway presents a self-signed leaf, so the wrong-listener case always fails as a clean transport error.

**Deploy order:** merge + deploy aceteam#5450 (coordinator dial, backward + forward compatible) first, then release this. In that order there is no window where SSH-deploy breaks. The gateway routes themselves (`/vnc`, `/modules/*`) need no coordinator change — this PR *restores* their intended `:8443` behavior.

## Test plan

| Group | Tests | Coverage |
| --- | --- | --- |
| `services/ports_test.go` (new) | 2 | registry membership + pairwise-distinct guard for every citadel-owned listener; control port ≠ gateway HTTPS port, clear of apps range + managed services |
| `internal/status/server_test.go` | +1 | `DefaultControlPort` regression guard (defaults off `services.GatewayHTTPSPort`, explicit override respected) |
| `cmd/gateway_bind_error_test.go` (new) | 2 | error message names port + dark routes; collision hint (control-port knobs) appears only for `listener already open` |
| existing | all | `go build ./...` clean; full `go test ./...` green, including the apps host-port collision guard now covering `ControlMTLSPort` |

Manual verification still worth doing on a canary node after release: `citadel work --gateway` on a mesh-connected node, confirm startup prints both `Gateway VPN: https://<mesh-ip>:8443` and `Control server VPN (mTLS): https://<mesh-ip>:8444`, and that `https://<mesh-ip>:8443/ping` (gateway) answers without a client cert.

## Related

- #504 (this fix), #501 (work-mode desktop parity, gated on this)
- aceteam-ai/aceteam#5419 (desktop relay re-point, gated on this), aceteam-ai/aceteam#5028 (control listener origin), aceteam-ai/aceteam#4709 (WhatsApp mesh dial broken by the collision)
- Companion coordinator PR: aceteam-ai/aceteam#5450

---
*Generated by Claude Fable 5*

### DIFF
--- a/cmd/gateway_bind_error.go
+++ b/cmd/gateway_bind_error.go
@@ -1,0 +1,34 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aceteam-ai/citadel-cli/services"
+)
+
+// gatewayVPNBindErrorMessage renders the startup error shown when the HTTPS
+// gateway cannot bind its mesh (tsnet) listener.
+//
+// This must be LOUD (#504): the gateway is the ONLY mesh entry point for
+// /vnc, /terminal, and /modules/*, so losing the bind means those routes are
+// unreachable over the mesh while the node still reports healthy in every
+// heartbeat. The old one-line "LAN-only" warning hid exactly that on every
+// node in the fleet after the mTLS control listener (#5028) started claiming
+// the same mesh :8443 first.
+//
+// tsnet reports a same-process double-bind as "listener already open" — that
+// is a collision between two citadel-owned listeners (a config/regression
+// bug, not an environment flake), so the message names the port and the
+// known culprit knobs instead of leaving the operator to guess.
+func gatewayVPNBindErrorMessage(port int, err error) string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "❌ GATEWAY MESH LISTENER FAILED: could not bind mesh port %d: %v\n", port, err)
+	b.WriteString("   Mesh routes /vnc, /terminal, and /modules/* are UNREACHABLE over the VPN.\n")
+	b.WriteString("   The gateway continues on LAN only; platform features that dial this node over the mesh (desktop, WhatsApp bridge, provisioned modules) will fail.\n")
+	if strings.Contains(err.Error(), "listener already open") {
+		fmt.Fprintf(&b, "   Another citadel listener already owns mesh port %d.", port)
+		fmt.Fprintf(&b, " Check --gateway-port against the mTLS control listener (CITADEL_CONTROL_PORT, default %d) and restart with distinct ports.", services.ControlMTLSPort)
+	}
+	return b.String()
+}
--- a/cmd/gateway_bind_error_test.go
+++ b/cmd/gateway_bind_error_test.go
@@ -1,0 +1,44 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/aceteam-ai/citadel-cli/services"
+)
+
+func TestGatewayVPNBindErrorMessageNamesPortAndRoutes(t *testing.T) {
+	msg := gatewayVPNBindErrorMessage(8443, errors.New("listen on 100.64.0.21:8443: connection refused"))
+
+	for _, want := range []string{"8443", "/vnc", "/terminal", "/modules/*", "LAN"} {
+		if !strings.Contains(msg, want) {
+			t.Errorf("message missing %q:\n%s", want, msg)
+		}
+	}
+	// Not a same-process double bind: no collision hint pointing at the control
+	// listener.
+	if strings.Contains(msg, "CITADEL_CONTROL_PORT") {
+		t.Errorf("collision hint should only appear for \"listener already open\" errors:\n%s", msg)
+	}
+}
+
+func TestGatewayVPNBindErrorMessageCollisionHint(t *testing.T) {
+	err := fmt.Errorf("listen on 100.64.0.21:8443: tsnet: listener already open for tcp, 100.64.0.21:8443")
+	msg := gatewayVPNBindErrorMessage(8443, err)
+
+	// A same-process collision must name the culprit knob and the control
+	// listener's current default so the operator can fix the config without
+	// spelunking (#504).
+	for _, want := range []string{
+		"CITADEL_CONTROL_PORT",
+		strconv.Itoa(services.ControlMTLSPort),
+		"--gateway-port",
+	} {
+		if !strings.Contains(msg, want) {
+			t.Errorf("collision message missing %q:\n%s", want, msg)
+		}
+	}
+}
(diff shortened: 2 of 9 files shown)


sunapi386:
Reopening after investigation: this was closed at 14:15Z with no comment, no explanation, code not on main, and #504 still open — 11 minutes before a batch of unrelated PR merges (#441/#425/#506/#431) from the same account. Best explanation is collateral from a parallel session's PR triage, not a decision. CI is green; founder asked for it to be investigated and the merge queue to proceed. NOTE: merging ≠ releasing — the next citadel release must wait until aceteam#5450 (coordinator dial fallback, merged) is DEPLOYED, per the rollout matrix in this PR.

---
*Generated by Claude Fable 5*

Issue #514, opened 2026-07-13
feat(meeting): package the meeting media stack as an installable citadel module (on-demand, containerized)


sunapi386:
## Context

The meeting/notetaker capability (aceteam#5097) is powerful but its "readiness" is fragile: a node is "meeting-capable" only if the **host** happens to have `google-chrome`, `Xvfb`, `pactl`/PulseAudio, and `ffmpeg` pre-installed. There is no way to *make* an arbitrary node a meeting node on demand — you either have the media stack on the box or you don't. Meanwhile the WhatsApp bridge solved exactly this class of problem by shipping as an **installable citadel module** (git repo of `service.yaml` + `compose.yml` → `docker compose up -d` → gateway route → reconciled desired state). The meeting stack should follow that precedent: package it as a module so meeting-capability becomes an **on-demand install** (via the MODULE_SET / desired-state path), not a host-provisioning prerequisite.

## The gap (one sentence)

The meeting media stack is hardcoded Go that shells out to host binaries, so meeting-capability can't be installed on demand the way every other provisioned service (e.g. WhatsApp) can.

## Grounding

**1. MEETING_JOIN runs in-process against HOST binaries — no container boundary.**
- `internal/worker/job.go:143` — `JobTypeMeetingJoin = "MEETING_JOIN"`.
- `internal/jobs/meeting_join.go` — `MeetingJoinHandler.Execute` orchestrates in-process: `platform.NewNullSinkRecorder` (`:331`) → `rec.LoadSink()` (`:332`, `pactl load-module module-null-sink`) → `platform.NewMeetingBrowser` (`:340`) → `br.Start()` (`:341`, host Chrome under Xvfb) → `rec.Start()` (`:357`, ffmpeg on the sink monitor).
- `internal/platform/meeting_browser.go:402` — the load-bearing line: `cmd.Env = withPulseSink(withDisplay(os.Environ(), display), b.sinkName)` composes `DISPLAY=<xvfb>` + `PULSE_SINK=<null sink>` onto the **host** browser process.
- `internal/platform/audio.go` — `NullSinkRecorder` shells `pactl` (`:166`) and `ffmpeg -f pulse -i <sink>.monitor -ac 1 -ar 16000` (`:93`, `:191`).

**2. Capability is auto-detected from host binaries (the fragility).**
[...]

PR #520, opened 2026-07-14
feat(catalog): self-healing resolution auto-refreshes a cold cache

### LINKED ISSUE #518: Self-healing catalog resolution: auto catalog-update when a module is missing from the local cache
## Context

Installing a catalog module on a node (`citadel module install <name>`, and the remote `MODULE_SET` / desired-state pull path) resolves the module from the node's **local catalog cache**. On a fresh node that cache is empty until `citadel service catalog update` is run manually — the install fails with "service '<name>' not found in catalog". This was hit for real during the meeting-module (#514) prod canary: the first `citadel module install meeting` failed until a manual `catalog update`. Over MCP there is **no way to trigger a catalog refresh remotely**, so an agent/operator cannot self-serve an install.

## Summary

Make catalog resolution **self-healing** in the install path:

- When a requested module is not found in the local catalog cache, automatically run the equivalent of `catalog update` (refresh/clone the configured catalog source(s)) **once**, then retry resolution before failing.
- Applies to both the interactive `citadel module install` and the `MODULE_SET` handler / reconcile-pull converge path, so a remote desired-state assignment installs cleanly on a node that has never run `catalog update`.
- Fail with the existing clear error only if the module is still absent after a refresh.

## Scope
- citadel-cli: `runCatalogInstall` (`cmd/catalog.go`), the `MODULE_SET` handler (`internal/worker/module_set.go`), and/or the reconcile `ModuleOps` adapter.
- Keep it idempotent and bounded (one refresh per resolution, not a retry storm).
- Test: install a module whose catalog cache is cold → succeeds after auto-refresh.

Requested by Jason.

sunapi386:
## Context

Installing a catalog module on a node — `citadel module install <name>`, and the remote `MODULE_SET` / desired-state reconcile-pull converge path — resolves the module from the node's **local catalog cache**. On a fresh node that cache is empty until someone runs `citadel service catalog update`, so the install fails with `service '<name>' not found in catalog`.

This isn't theoretical: it bit the meeting-module (#514) prod canary — the first `citadel module install meeting` failed until a manual `catalog update`. And over MCP / remote desired-state there's **no operator** to run `catalog update`, so an agent-driven assignment could never converge on a cold node.

Closes #518.

## Summary

Make catalog resolution **self-healing** in the install path: when a requested module is absent from the local cache, automatically refresh the configured source(s) **once**, then retry before failing.

- **`ResolveCatalogServiceSelfHealing`** (`internal/catalog/selfheal.go`) wraps the atomic single-source resolver used by `runCatalogInstall` (interactive install). The atomic resolution stays intact — this is the shadow-safety invariant (a community source can't shadow a default service's name to install its own compose under default privileges), so the wrapper sits *around* it, not through `LoadServiceManifest`.
- **`LoadServiceManifestSelfHealing`** wraps the catalog branch of `resolveModuleForTUI`, which covers the `MODULE_SET` handler, the reconcile-pull converge path (`liveModuleOps.Install`), **and** the TUI install callbacks in one edit. The follow-up `GetComposeFile` reads the now-warm cache, so a single refresh covers both manifest and compose.
- **Bounded**: exactly one refresh per resolution attempt — no retry storm.
- **Partial-refresh-tolerant** (the subtle part): `Update()` refreshes every source and returns the *first* error even when the rest — including the built-in default — succeeded. So a node with a broken community source added must still install a default-catalog module like `meeting`. The wrapper retries resolution first and only surfaces the refresh error if the module is *still* absent. This mirrors `Update()`'s own "one broken repo never blocks the rest" contract.
- A new `ErrServiceNotFound` sentinel (wrapped at both not-found sites) scopes the refresh to genuine cache misses; other resolution errors (unparseable manifest, stat failure) return immediately without a refresh. The existing clear `not found in catalog` message is preserved.

## Test plan

New `internal/catalog/selfheal_test.go` (mocks the refresh via an injectable package var — no git/network), mirroring the package's `seedService` / `t.Setenv("HOME", …)` patterns:

| Test | Covers |
|------|--------|
| cold cache → refresh seeds cache → resolves | happy path; asserts refresh ran **exactly once** (bounded) |
| still absent after refresh | returns the clear `not found in catalog` error; refresh ran once |
| refresh returns a partial error but module resolves anyway | the landmine guard — install must not be refused on a partial `Update()` error |
| warm cache | resolves with **zero** refreshes (no needless network cost) |
| `LoadServiceManifestSelfHealing` cold cache | the reconcile / MODULE_SET path self-heals with one refresh |

`go build ./...`, `go vet`, and `go test ./...` all green.

## Needs live-node verification

- On a genuinely cold node, `citadel module install meeting` should now succeed without a prior `catalog update` (the #514 repro).
- A remote `node_module_set` → `MODULE_SET` assignment for a default-catalog module on a never-updated node should converge cleanly.

Opening as draft pending that on-node check.

## Related

- #518 (this) · #514 (meeting-module canary that surfaced it)

### DIFF
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -316,7 +316,7 @@ func runCatalogInstall(cmd *cobra.Command, args []string) error {
 	// atomically. This is a security invariant: the trust decision must key the
 	// exact compose that gets installed, so a community source cannot shadow a
 	// default service's name to install its own compose under default privileges.
-	resolved, err := catalog.ResolveCatalogService(name)
+	resolved, err := catalog.ResolveCatalogServiceSelfHealing(name)
 	if err != nil {
 		return err
 	}
--- a/cmd/module.go
+++ b/cmd/module.go
@@ -799,7 +799,10 @@ func buildModuleInstallCallbacks() controlcenter.ModuleInstallCallbacks {
 // external source it clones/updates the repo and returns the full ResolvedModule.
 func resolveModuleForTUI(src catalog.Source) (manifest *catalog.ServiceManifest, composeSrc string, resolved *catalog.ResolvedModule, err error) {
 	if src.Kind == catalog.KindCatalog {
-		manifest, err = catalog.LoadServiceManifest(src.Name)
+		// Self-healing: on a cold catalog cache this refreshes the configured
+		// source(s) once and retries, so a remote MODULE_SET / reconcile install
+		// (and the TUI install) succeeds on a node that never ran `catalog update`.
+		manifest, err = catalog.LoadServiceManifestSelfHealing(src.Name)
 		if err != nil {
 			return nil, "", nil, err
 		}
--- a/internal/catalog/catalog.go
+++ b/internal/catalog/catalog.go
@@ -405,7 +405,7 @@ func LoadServiceManifest(name string) (*ServiceManifest, error) {
 			return nil, err
 		}
 	}
-	return nil, fmt.Errorf("service '%s' not found in catalog", name)
+	return nil, fmt.Errorf("service '%s' not found in catalog: %w", name, ErrServiceNotFound)
 }
 
 // loadManifestFromPath reads a service's service.yaml from a single source's
--- a/internal/catalog/selfheal.go
+++ b/internal/catalog/selfheal.go
@@ -1,0 +1,87 @@
+// internal/catalog/selfheal.go
+//
+// Self-healing catalog resolution (aceteam-ai/aceteam#518).
+//
+// The install path resolves a module from the node's LOCAL catalog cache. On a
+// COLD cache (a fresh node that never ran `citadel service catalog update`) that
+// resolution fails with "service '<name>' not found in catalog" -- exactly what
+// bit the meeting-module (#514) prod canary. Over the remote MODULE_SET /
+// reconcile-pull converge path there is no operator to run `catalog update`
+// manually, so a desired-state assignment could never install on a cold node.
+//
+// These wrappers make resolution self-healing: when a service is absent from the
+// local cache, refresh the configured catalog source(s) ONCE, then retry before
+// failing. Bounded to a single refresh per resolution (no retry storm); a service
+// still absent after the refresh returns the same clear error.
+package catalog
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrServiceNotFound is the sentinel wrapped by the catalog resolvers when a
+// service name is absent from every configured source's LOCAL cache. The
+// self-healing wrappers key their single catalog refresh on this sentinel: a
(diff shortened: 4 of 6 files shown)


sunapi386:
Revived and validated after 6 weeks stale. Summary:

**Rebase:** clean, conflict-free rebase onto current `main` (f33a2d7, includes #830's lockfile-scoped `ListInstalled`). The diff content vs `main` is byte-identical to the original PR — no manual conflict resolution needed. `recordCatalogModuleLock` (#830 follow-up) sits in a different part of `runCatalogInstall` than this PR's `ResolveCatalogServiceSelfHealing` call, so the two changes compose cleanly.

**Relevance check:** still fully relevant, not superseded. `resolveModuleForTUI` (wrapped here via `LoadServiceManifestSelfHealing`) is still the single resolution point for the `MODULE_SET` handler (`cmd/module_ops.go:110`), the reconcile-pull converge path, and the TUI install callbacks — unchanged since July. `Update()`'s partial-failure-tolerant contract (first error returned even when other sources, including the default, succeeded) is also unchanged, so the wrapper's "retry resolution before surfacing the refresh error" logic still matches reality.

**Verification beyond the original test suite:**
- Live cold-cache smoke test (not committed — throwaway): `ResolveCatalogServiceSelfHealing("meeting")` against a genuinely empty `$HOME`, exercising the REAL `Update()` (actual git clone of `github.com/aceteam-ai/citadel-services`), not the stubbed refresh the unit tests use. Resolved successfully in 0.71s — confirms the PR's core claim end-to-end.
- Added one more unit test, `TestResolveSelfHealingStillAbsentAndRefreshFailed` (committed), covering the previously-untested branch where the refresh call itself errors (offline node) AND the module is still absent — pins that the surfaced error mentions both the not-found and the refresh-failure text.
- Checked amplification risk: a persistently-unresolvable module name (typo/deleted catalog entry) now costs a full catalog `git pull` per reconcile attempt instead of a cheap local miss. This is bounded by the existing reconcile loop's cadence (5-min ticker, 10s nudge debounce, unchanged by this PR) rather than a new unbounded loop, and is a reasonable trade-off given the alternative (permanent unrecoverable failure with no operator to intervene) — noting it here rather than blocking on it.
- Checked TUI-freeze risk: the non-catalog branch of `resolveModuleForTUI` (`catalog.ResolveSource`) already performs an unconditional synchronous git clone for every external-source TUI install today. This PR only extends that same pre-existing blocking-call tolerance to the catalog branch, and only on a cold-cache miss — not a new hazard class.

`go build ./...`, `go vet ./...`, `gofmt -l .` (empty), and `go test ./...` (full suite, all packages) all green on the rebased branch.

Marked ready for review and force-pushed the rebased branch.

PR #561, opened 2026-07-17
gate :8080 control endpoints behind org token (#5028 lever B)

sunapi386:
## Summary

Hardens authentication on the node `:8080` status/control server. Control endpoints (`/provision/*`, `/workflow/run`, `/agent/*`) previously trusted callers by source-IP range; this PR adds the ability to require a per-org bearer token for those endpoints instead, as defense-in-depth. Part of the `aceteam-ai/aceteam#5028` hardening track.

## Approach

- Reuses the existing server-decryptable **per-org token** already validated on the `:7860` listener via `terminal.TokenValidator.ValidateToken(token, orgID)` — no new key material or secret store. Presented as `Authorization: Bearer <token>`.
- The token authenticates "a valid token for this node's own org." Durable per-caller identity (coordinator mTLS, #5959) is the follow-up; the SSH-key route already runs on that mTLS listener.

## What changed

- `internal/status/ssh_handler.go` — when enforcement is on, the per-org token becomes mandatory. Adds `RequireControlTokenEnvVar` + `RequireControlTokenEnabled()` (truthy set `1/true/yes/on`, matching the `autostop.go` pattern).
- `internal/status/server.go` — `RequireControlToken` field on `Server` + `ServerConfig`.
- `cmd/work.go` — wires `RequireControlToken` from `CITADEL_REQUIRE_CONTROL_TOKEN`; logs the posture and warns loudly if enforcement is on but no validator is configured (so the fail-closed state is not mistaken for a bug).
- `internal/status/ssh_handler_test.go` — `TestRequireControlTokenEnforcement`: unauth / wrong-token / cross-org-token requests rejected (401); correct org token accepted (200); default-off path preserves existing behavior.

## Rollout (breaking change, so gated)

Default is **OFF** so existing relays keep working at merge. Enable per node via `CITADEL_REQUIRE_CONTROL_TOKEN=true` **after** the relay is updated to present the bearer, in a planned window. Rollback is code-free (unset env + restart). With enforcement on but no validator configured, control endpoints fail closed.

This is the deliberate exception to no-feature-flags-by-default: a breaking-change kill-switch, not a feature flag.

## Testing

- `go build ./...` — clean.
- `go vet ./internal/status/ ./cmd/` — clean.
- `go test ./internal/status/` — pass, including the new enforcement test.

### DIFF
--- a/cmd/work.go
+++ b/cmd/work.go
@@ -1183,6 +1183,13 @@ func runWork(cmd *cobra.Command, args []string) {
 			Port:    workStatusPort,
 			Version: Version,
 			Agent:   agentProviders,
+			// Interim bearer-token enforcement on the :8080 control endpoints
+			// (#5028 lever B). When set, the mesh-origin bypass in requireVPNOrAuth
+			// is dropped and the per-org terminal token is mandatory even for mesh
+			// peers, closing cross-org control access on the flat mesh. Default OFF
+			// so older relays that dial over the mesh without a bearer keep working;
+			// flip it on in a planned window once the relay presents the token.
+			RequireControlToken: status.RequireControlTokenEnabled(),
 		}
 
 		// Publish the gateway's self-signed leaf cert at GET /gateway-cert.pem so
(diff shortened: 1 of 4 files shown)


sunapi386:
Closing to keep the interim :8080 control-token gate out of the public PR stream until the planned #5028 enforcement window. Branch fix/5028-control-token-gate is preserved and will be reopened + merged at flip time.

PR #596, opened 2026-07-23
Default-off + passcode for console/screen/file capabilities (#6524)

sunapi386:
## Context

Part of aceteam-ai/aceteam#6524. On the White Whale onboarding call, Peter saw the web console showing his MacBook's terminal/screen/files as a live node the moment it joined the fabric — a trust-breaking default for a privacy-first customer. Node console/desktop/file surfaces were **default-ON**; this flips the three sensitive remote-access surfaces to **default-OFF opt-in + a per-node passcode** (Jason's decision: both layers), while inference/model-serving stays available with zero remote access enabled.

These three already existed as `Permissions{Console, Desktop, Files}` (default `true`), gated at the HTTPS gateway. Rather than add parallel `console.go`/`desktop.go`/`files.go` files, this flips those defaults and adds the passcode to the existing struct — the `Default<Cap>()`-returns-DISABLED shape the issue asked for.

## Summary

**Default-OFF (`internal/config/permissions.go`)**
- `DefaultPermissions()` now returns `console/desktop/files` (and pre-existing `shell`) **disabled**; `services/ssh/provision` stay on. Absent keys in a legacy `permissions.yaml` stay disabled (fail closed).

**Per-node passcode (bcrypt, `golang.org/x/crypto`)**
- `PasscodeHash` (bcrypt, salt embedded — no plaintext, no separate salt) with `SetPasscode` / `VerifyPasscode` / `HasPasscode`. `permissions.yaml` written `0600` (now holds a credential). Fail-closed: an enabled surface with no passcode set stays denied.

**Enforcement at every path a fresh node could leak from (each fails closed):**
1. **Advertisement** (`internal/status/capabilities_flags.go`) — heartbeat console/desktop/files flags now require capability AND permission, so the web console stops presenting a live surface for a fresh node. Enablement stays separately visible via the `PermissionState` heartbeat block so the web can render an enable-CTA. GPU/H264 not gated (inference advertises regardless).
2. **Redis fabric handlers** (`internal/worker/handler_adapter.go`) — VNC/screenshot/actions and FILE_* browse/host handlers register only when `desktop`/`files` are enabled (`DesktopDisabled`/`FilesDisabled`). A fresh node refuses those jobs. `TRANSCRIBE_AUDIO`/`MEETING_JOIN` share the workspace but are the default-ON meeting capability and are deliberately **not** gated.
3. **HTTPS gateway** (`internal/gateway/gateway.go`) — `permissionMiddleware` blocks a disabled category and, for an enabled sensitive category, additionally requires the passcode (`X-Citadel-Passcode` header / `?passcode=`).
4. **Direct mesh listeners** (the key non-gateway threat — a mesh peer can dial these directly):
   - Terminal server started only when `console` is enabled (`cmd/work.go`), and every connection additionally requires the passcode (`terminal.Config.PasscodeVerifier`) on top of token/mesh-identity auth.
   - Status-server desktop endpoints (`/api/screenshot`, `/api/actions`) require the passcode in `requireAuth` (`status.ServerConfig.PasscodeVerifier`).

**Opt-in paths**
- `APPLY_DEVICE_CONFIG` (`internal/jobs/config_handler.go`): new `consoleEnabled`/`desktopEnabled`/`filesEnabled` (`*bool`, nil = untouched) and `nodePasscode` (`*string`, bcrypt-hashed before persist). Writes the same `permissions.yaml` the gates read.
- Control Center Built-in Services modal: detail copy updated to the new posture; enabling a surface with no passcode set warns that access stays denied until a passcode is set.

**Docs:** `docs/capability-toggles.md` rewritten to document the default-OFF + passcode posture and contrast it with default-ON meeting/telemetry and **why** they differ.

## Test plan

`go build ./...`, `go vet`, and `go test ./...` all pass. New tests:
- `internal/config/permissions_test.go` — fresh node has the three surfaces off; absent keys stay disabled; passcode hashes (not plaintext), verifies, fails closed on unset/empty/cleared; `0600` file mode; round-trips through save/load.
- `internal/worker/handler_gating_test.go` — fresh node refuses VNC/FILE jobs; **inference + meeting/transcribe/TTS served regardless** (proves inference isn't gated); enabling restores the handlers.
- `internal/gateway/passcode_test.go` — disabled→403; enabled-no-passcode→401; wrong→401; correct→200; non-sensitive route needs no passcode.
- `internal/status/advertisement_gate_test.go` — capable-but-disabled node does not advertise (even with a live VNC port + workspace); enabling files re-advertises.
- `internal/jobs/config_handler_passcode_test.go` — APPLY_DEVICE_CONFIG enables console + sets passcode without writing plaintext.

## Design decisions to confirm (@sunapi386)

1. **Passcode scope = per-node, per-capability enablement** (the issue's recommendation): enable console/desktop independently; one node passcode unlocks whatever is enabled. Change if you want per-capability passcodes.
2. **Files has no direct-mesh HTTP surface** — file browse/host is only the Redis FILE_* path (org-backend-authenticated) plus the (routeless) gateway `files` category. So files is enforced by **default-OFF enablement + org-backend auth**, not a per-request passcode (there's no node HTTP file listener to attach one to). Console and desktop, which DO have direct mesh listeners, get the full passcode gate. Flag if you want files to also require a passcode on a future file HTTP surface.
3. **Breaking change for deployed nodes:** flipping the defaults means any already-deployed node relying on default-on console/desktop/files (including the computer-use / shared-desktop fleet) stops responding to those until re-enabled + a passcode is set. Intended per the issue, but calling it out.
4. **Applies-on-restart for the gateway path:** the gateway snapshots permissions at startup (existing model), so a passcode/enable pushed via APPLY_DEVICE_CONFIG takes effect for the gateway on next worker start. The terminal/desktop verifiers load fresh per request (honor rotation live). Both directions fail closed.

**Cross-repo follow-up (aceteam, NOT in this PR):** the web console should render these as opt-in ("enable + set passcode") instead of a live console, and send the passcode via `X-Citadel-Passcode` (or `?passcode=` for the WS terminal/VNC). Coordinate with mesh-ACL work (#5028): that's cross-org exposure, this is per-node owner consent.

### DIFF
--- a/cmd/controlcenter.go
+++ b/cmd/controlcenter.go
@@ -2005,6 +2005,8 @@ func runTUIWorker(ctx context.Context, activityFn func(level, msg string)) error
 		ConfigDir:                 ccConfigDir,
 		AllowReadOutsideWorkspace: resolveAllowReadOutsideWorkspace(),
 		ShellDisabled:             !ccPerms.Shell,
+		DesktopDisabled:           !ccPerms.Desktop,
+		FilesDisabled:             !ccPerms.Files,
 		WorkflowExec:              ccWfExec,
 		HandlerLog:                func(format string, args ...any) { activity("info", fmt.Sprintf(format, args...)) },
 	}
--- a/cmd/controlcenter.go
+++ b/cmd/controlcenter.go
@@ -517,10 +517,20 @@ func ccOnNetworkConnect(activityFn func(level, msg string)) {
 			stopTerminalServer()
 		}
 
-		if err := startTerminalServer(orgID, activityFn); err != nil {
-			activityFn("warning", fmt.Sprintf("Terminal server failed: %v", err))
+		// Console permission gate (aceteam#6524): the control-center node stands up
+		// its own terminal server with a VPN listener, so — exactly like `citadel
+		// work` — starting it must be gated on the operator opting `console` in, not
+		// just on being connected. A fresh node has `console` default-DENY, so no
+		// terminal listener goes on the org mesh. When enabled, the passcode
+		// verifier wired in startTerminalServer gates every connection.
+		if config.LoadPermissions(platform.ConfigDir()).Console {
+			if err := startTerminalServer(orgID, activityFn); err != nil {
+				activityFn("warning", fmt.Sprintf("Terminal server failed: %v", err))
+			} else {
+				activityFn("info", fmt.Sprintf("Terminal server listening on port %d", ccTerminalPort))
+			}
 		} else {
-			activityFn("info", fmt.Sprintf("Terminal server listening on port %d", ccTerminalPort))
+			activityFn("info", "Console (terminal) disabled — enable the 'console' permission + set a node passcode to allow remote terminal access")
 		}
 	}
 
@@ -562,10 +572,20 @@ func startTerminalServer(orgID string, activityFn func(level, msg string)) error
 	}
 
 	// Build configuration
-	config := terminal.DefaultConfig()
-	config.OrgID = orgID
-	config.AuthServiceURL = authServiceURL
-	config.Port = ccTerminalPort
+	termCfg := terminal.DefaultConfig()
+	termCfg.OrgID = orgID
+	termCfg.AuthServiceURL = authServiceURL
+	termCfg.Port = ccTerminalPort
+
+	// Per-node passcode gate (aceteam#6524), same as the `citadel work` terminal
+	// server: even after a valid token or same-owner mesh identity, a connection
+	// must present the node passcode. Loaded fresh per connection so a rotated
+	// passcode is honored without restarting the control center; fails closed.
+	// (config is the internal/config package; the local terminal config is
+	// termCfg to avoid shadowing it here.)
+	termCfg.PasscodeVerifier = func(pin string) bool {
+		return config.LoadPermissions(platform.ConfigDir()).VerifyPasscode(pin)
+	}
 
 	// Create the caching token validator
 	apiToken := ""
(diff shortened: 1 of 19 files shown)


sunapi386:
### Follow-up audit: other server-construction sites

Audited every site that stands up a sensitive listener (the passcode verifier is nil-able, so each site must wire it):

- **`cmd/work.go`** — terminal (console gate + passcode), status desktop endpoints (passcode). ✅
- **`cmd/serve.go`** — only the gateway (`SetPermissions` carries the passcode, so `/vnc`,`/terminal` are gated); no independent terminal/desktop listener. ✅
- **`cmd/controlcenter.go`** — a control-center-only node stands up its **own** terminal server with a VPN listener. It was ungated on `console` and had no passcode → **fixed** in the latest commit: start gated on `console`, and `terminal.Config.PasscodeVerifier` wired (loads perms fresh per connection). ✅

**Residual (design decision #5, for @sunapi386):** the control-center desktop path is the **native VNC + H.264 deskstream** servers (`startVNCServer`/`startH264Server`), not the status-server `/api/screenshot` endpoints. They attach VPN listeners and are gated on `desktop` (**default-OFF**, so a fresh node exposes nothing — the White Whale scenario is closed), but they have **no per-connection auth hook**, so the new node passcode is not yet enforced on that specific stream when an operator enables `desktop` on a control-center node. Node-passcode-gating those requires adding an auth handshake to `desktop.VNCServer` / `deskstream.Server` — a separate change to those servers, tracked as a follow-up. The default-OFF flip is the load-bearing protection here; the passcode fully covers console (both `work` and control-center), the gateway `/vnc`+`/terminal`, and the status-server desktop endpoints.

**Also note:** the terminal `PasscodeVerifier` is set unconditionally, so it now applies to the **localhost/LAN** bind too, not just the mesh — any local `citadel connect`/tooling that previously needed only a token now also needs the node passcode. Intended (defense in depth), flagging so it is not a surprise.

Issue #597, opened 2026-07-24
Add nvr catalog module — self-hosted camera NVR (wyze-bridge + Frigate)


sunapi386:
## Summary

Package the hand-built NVR stack (docker-wyze-bridge + Frigate) as an assignable Citadel catalog module (`nvr`), so any node can be given camera recording + local object detection via `fabric_node_module_set module=nvr`. A reference implementation is running **today** on node **1314 (`mini`, SJC)** recording to a Synology NFS share — this issue is about turning that hand-built instance into a catalog SKU.

Companion issue: **Ingress / fabric exposure** (separate issue — independent of this one; the NVR only needs to *record*, exposure is handled there).

## Decisions (locked)

- First deliverable = a real **catalog module**, not a one-off compose bundle.
- Storage is configurable per-assignment: `local | nas | volume`.
- Independent of the Ingress work — do not couple them.

## Where it goes

- `internal/catalog/` — catalog entry + install/lifecycle/runtime (`catalog.go`, `install.go`, `lifecycle.go`, `runtime.go`, `gpu_compose.go`). Model it on how `ollama`/`llamacpp`/`bonsai` are declared.
- `internal/apps/` — docker/compose app definition (`docker.go`, `state.go`; mind the existing `container_name_collision` / `hostport_collision` handling — host networking interacts with these).
- `internal/mesh/discovery.go` — the known-service list where `ollama`/`llamacpp`/`bonsai` appear; add `nvr`/`frigate` if discovery metadata belongs here (exposure itself is the companion issue).

## Module spec

Two containers, orchestrated via compose:

1. **wyze-bridge** (`idisposablegithub365/wyze-bridge:latest`) — **MUST run host networking.** TUTK P2P needs LAN broadcast + UDP hole-punching; Docker bridge NAT breaks it and *every* camera fails with `wyze: connect failed: discovery timeout`. Non-negotiable, and the #1 gotcha.
2. **frigate** (`ghcr.io/blakeblackshear/frigate:stable`, currently 0.17.x) — recording + detection. Needs `/dev/dri/renderD128` mapped for hardware decode.

### Config schema (assignment input)

```yaml
nvr:
  wyze: { email, password, api_id, api_key }   # secrets: from assignment, never logged
  retention_days: 12
  detector: openvino | cpu     # openvino = Intel iGPU (device: GPU); cpu fallback for AMD (Vega has no OpenVINO)
  storage:
    mode: local | nas | volume
[...]

PR #605, opened 2026-07-25
fix(nvr): make nvr-config image build (patches/ + buildvcs)

sunapi386:
## Why
`build-nvr-service` is **red on main** (run 30059248662) — it runs post-merge, not as a required check, so #600 merged with a broken image build. Result: `ghcr.io/aceteam-ai/nvr-config` was never pushed (404), and `fabric_node_module_set module=nvr` cannot come up — Frigate blocks forever on the missing init container.

## Two bugs (both proven by a local `docker build` on ubuntu-gpu)
1. **`patches/` missing before `go mod download`.** `go.mod` has `replace github.com/gdamore/tcell/v2 => ./patches/tcell/v2`; only `go.mod`/`go.sum` were copied, so download can't read the replaced module. → `COPY patches/ ./patches/` before the download layer (kept split so the module layer still caches).
2. **`go build` VCS stamping fails** (`error obtaining VCS status: exit status 128`) — needs a usable `.git` in-context. → `-buildvcs=false`, standard for containerized Go builds.

## Verified
Image builds (11.2MB); binary runs and fails loudly on missing `NVR_CAMERAS` as designed. Once merged, CI publishes the image and `module=nvr` can deploy to node 1314.

Refs #597, #600

### DIFF
--- a/services/nvr-service/Dockerfile
+++ b/services/nvr-service/Dockerfile
@@ -10,9 +10,10 @@
 FROM golang:1.26 AS build
 WORKDIR /src
 COPY go.mod go.sum ./
+COPY patches/ ./patches/
 RUN go mod download
 COPY . .
-RUN CGO_ENABLED=0 GOOS=linux go build -trimpath -o /out/nvrconfig ./cmd/nvrconfig
+RUN CGO_ENABLED=0 GOOS=linux go build -trimpath -buildvcs=false -o /out/nvrconfig ./cmd/nvrconfig
 
 FROM alpine:3.20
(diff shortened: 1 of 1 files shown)


sunapi386:
Superseded by #615, which already landed both fixes on main (COPY patches/ before go mod download, and -buildvcs=false) and whose build-nvr-service run (30306267160) succeeded and published ghcr.io/aceteam-ai/nvr-config. This branch now conflicts precisely because main already has the fix. Closing as redundant.

Issue #616, opened 2026-07-27
feat(embedding): ship TEI as a first-class Citadel module — mesh-exposed on :8102 + heartbeat-advertised so sovereign embeddings route


sunapi386:
## Context

Sovereign RAG embeddings shipped to prod (`aceteam-ai/aceteam#6339` + `aceteam-ai/aceteam#6495`, live on `production` @ v1.335.0). The backend now resolves each org's embedding model and, for a `TEI` model, embeds on the org's **own Citadel node** instead of OpenAI — storing vectors with model provenance and searching via `match_documents_scoped_v2`. The whole storage/search/provenance half is verified working against live prod (see "What's already validated" below).

**The one thing that does not work yet is the actual node hop.** The backend can't reach a node's TEI, because TEI is not a first-class Citadel module — it only exists as a hand-run `docker run` container, which lands on the wrong side of the network fence and is invisible to service discovery. This issue is to ship TEI as a proper module so the prod path routes end-to-end.

## Root cause (both proven on `ubuntu-gpu-3` today)

The backend resolver (`aceteam` `python-backend/utils/rag_embeddings.py`) does two things a manual container can't satisfy:

1. **Mesh reachability** — `resolve_node_tei_base_url()` (L248) returns `http://{node_headscale_ip}:8102/v1` and the backend hits it over the tailscale SOCKS5 relay. But a manual `docker run -p 8102` binds the **host kernel** tailscale IP, while Headscale has the node registered under Citadel's **userspace** tailscaled IP — different network namespace:

   | Endpoint | Result |
   |---|---|
   | `100.64.0.182:8102` (host kernel TS — where a manual container binds) | **HTTP 200** |
   | `100.64.0.78:8102` (Citadel userspace TS — what the backend actually resolves + hits) | **connection refused** |

2. **Service discovery** — `_find_tei_node()` (L205) only routes to an online org node whose heartbeat advertises a service name matching `_TEI_SERVICE_MARKERS` (L66: `tei`, `text-embeddings-inference`, `text_embeddings_inference`, `embedding`), read via `_get_node_services()` in `python-backend/routes/fabric_models.py` (L218: Redis `node:service_usage:{node_id}` → persisted `fabric_node_status.service_usage` → `running_services`). A hand-run container is not a Citadel-managed module, so it never appears in any of those — **no online node advertises a TEI service today**, so discovery finds nothing even when TEI is up.

## Proposed plan

[...]

PR #621, opened 2026-07-28
feat(embedding): ship TEI as a first-class module so sovereign embeddings route (#616)

sunapi386:
## Context

Closes the citadel half of **aceteam-ai/citadel-cli#616**. Sovereign RAG embeddings (aceteam #6339/#6495, live on prod) let an org embed on its **own** Citadel node's TEI instead of OpenAI. The backend was shipped but the path never routed. Root cause, proven on a live node:

- **Netns split:** a hand-run TEI container binds the *host-kernel* tailnet, but the backend resolves the node's *citadel userspace* tailnet IP — different network namespaces, so a dial to `:8102` is refused. (The paired backend PR fixes transport by routing through the gateway, which already mesh-proxies `/v1/embeddings`.)
- **Discovery gap:** the heartbeat collector advertises services via the **engine probe** (its manifest `c.services` is nil in the heartbeat path), so nothing surfaced a running TEI. `_find_tei_node` matched no node.

## Summary

- **`services/compose/tei.yml`** — embedded `tei` module: `text-embeddings-inference:cpu-1.6` serving `gte-multilingual-base` (768-dim) on the OpenAI-compatible `/v1/embeddings`, host-published on the fixed `8102` the gateway proxies to. `--dtype float32` + single-thread `MKL/OMP/RAYON` is **load-bearing** (float16-on-CPU segfaults, citadel-services#14) and pinned by a compose-contract test. Registered in `ServiceMap` (auto-materializes + advertised as deployable).
- **Embedding discovery probe** (`internal/status/engines.go`) — a purpose-built `collectEmbeddingServiceStatus` advertises a running `citadel-tei` container, gated on TEI `/health`, as `ServiceInfo{Type: "embedding"}`. Kept **separate from the LLM engine probe** so an embedding server is never fed to the gateway chat router. This is the signal `_find_tei_node` matches, and it survives the backend's heartbeat→`service_usage` pass-through unchanged (a model-less `type:"embedding"` entry).
- **Port ownership refactor** — TEI (`8102`) moves from a `ReservedCitadelPorts` standin (it predated TEI being a module) to a **compose-owned** port, mirroring the transcribe sidecar (8101). Because a fixed compose port is deliberately *not* in `ReservedCitadelPorts` (the collision guard forbids a module publishing a reserved port), a new **`ClaimedHostPorts()`** union (reserved + fixed compose ports) is now consulted by **both** dynamic allocators — app auto-allocation and the whatsapp bridge — so neither races transcribe/tei for their ports. (The whatsapp allocator previously read `ReservedCitadelPorts` directly and could have grabbed 8101/8102.)

## Test plan

- `go test ./services/... ./internal/status/... ./internal/apps/... ./internal/whatsapp/...` green; `go vet` clean.
- New guards: TEI compose contract (float32/MKL/port/image), `GetAvailableServices` includes tei, probe port resolution + `ServiceTypeEmbedding != ServiceTypeLLM`, `ClaimedHostPorts` covers transcribe+tei and keeps tei out of `ReservedCitadelPorts`.
- **Verified on a live node:** `citadel-tei` running + TEI `/health` 200 satisfy the probe's exact conditions.

**Draft** — final acceptance is deploy-gated: roll this binary to a node, confirm it advertises `tei`, and that the backend auto-discovers it (paired backend PR). Pairs with the aceteam backend PR for the transport half.

### DIFF
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -12,21 +12,21 @@ import (
 // accepts both embedded/catalog services (services.ServiceMap) and
 // module-installed services tracked in the manifest, while rejecting unknown
 // names. This is the regression guard for issue #358, where the gate only
-// consulted ServiceMap and rejected module-installed services like "tei".
+// consulted ServiceMap and rejected module-installed services not in the embedded ServiceMap.
 func TestServiceIsKnown(t *testing.T) {
-	// "tei" is a stand-in for a module-installed service: it is NOT in
+	// "custommod" is a stand-in for a module-installed service: it is NOT in
 	// services.ServiceMap but is present in the manifest after
-	// `citadel module install tei`.
+	// `citadel module install custommod`.
 	manifest := &CitadelManifest{
 		Services: []Service{
-			{Name: "tei", ComposeFile: "./services/tei.yml"},
+			{Name: "custommod", ComposeFile: "./services/custommod.yml"},
 		},
 	}
 
 	// Sanity check: the module-installed name must not be an embedded service,
 	// otherwise case (b) would not actually exercise the manifest fallback.
-	if _, ok := services.ServiceMap["tei"]; ok {
-		t.Fatal("test assumption broken: 'tei' should not be in services.ServiceMap")
+	if _, ok := services.ServiceMap["custommod"]; ok {
+		t.Fatal("test assumption broken: 'custommod' should not be in services.ServiceMap")
 	}
 
 	tests := []struct {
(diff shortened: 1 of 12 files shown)


sunapi386:
Paired backend PR (transport half): **aceteam-ai/aceteam#6704** — points `resolve_node_tei_base_url` at the gateway's `/v1/embeddings` and skips TLS verify for the per-node self-signed cert. Land + deploy both, then run the live E2E.

Issue #637, opened 2026-07-30
nvr module: raise shm_size to 512mb and add a node-local MQTT broker for event egress


sunapi386:
## Summary

Two follow-ups to the `nvr` module (#597), both found while running it for real on node **1314 (`mini`, SJC)** — it is currently recording 3 cameras to a Synology NFS share.

1. **Raise `shm_size`** — the shipped 256 MB is genuinely too small.
2. **Add a node-local MQTT broker + event forwarding** — Frigate has no other real-time egress, and this is the prerequisite for native push notifications.

## 1. shm_size: 256mb -> 512mb

Frigate logged, on startup with 3x 1080p cameras:

```
frigate.camera.maintainer WARNING : The current SHM size of 256.0MB is too small, recommend increasing it to at least 306MB.
```

This is not cosmetic. `/dev/shm` is where Frigate passes **raw decoded frames** between the capture and detect processes (a 1080p YUV420 frame is ~3 MB, buffered ~9 deep per camera). After bumping to 512 MB, measured steady-state usage is:

```
shm  512M  322M used  63%  /dev/shm
```

**322 MB in use — above the old 256 MB ceiling.** Undersizing here causes frame-write failures and capture-process restart loops, not just slowness. Note `shm_size` is a cap on a tmpfs, not a reservation, so raising it costs no RAM until used.

Change `services/nvr-service/compose.yml`: `shm_size: 256mb` -> `512mb`.

## 2. Node-local MQTT broker + event forwarding

**Why MQTT at all:** Frigate 0.17 has **no generic outbound webhook**. Confirmed against the running instance's config schema — the only relevant sections are `MqttConfig`, `NotificationConfig` (that is *web push*, which needs HTTPS/service-worker and is useless for native APNs), and `AlertsConfig`. So the only real-time egress for events is MQTT; the alternative is polling `/api/events`, which is laggy and forces last-seen-id bookkeeping.

**Why node-local rather than one shared broker for the fabric** (this was considered and rejected):

- Frigate publishes at **QoS 0** (verified in the live config: `"qos": 0`). Fire-and-forget — any broker blip silently drops events. A loopback broker makes that window ~zero; a WAN broker makes it real.
- A shared bus is a **second control plane** to secure (TLS, credentials, ACLs) when the citadel worker already holds an authenticated, outbound-only channel to the fabric.
[...]

PR #646, opened 2026-07-31
fix(catalog): node-generated module secrets — unbreak nvr assignment/update (#637)

sunapi386:
## Context

#639 gave the nvr module a node-local MQTT broker and made `NVR_MQTT_PASSWORD`
a required config var so mosquitto could never start anonymously. The compose
`:?` guard was right; making the operator the only possible SOURCE of the value
was not.

Nothing produces that value. Grep on main finds only consumers
(`cmd/nvrconfig`, the compose) and the test that asserts the `:?`. So
`cmd/module_ops.go:163` — the fabric MODULE_SET assignment path — resolves
config with `interactive=false` and the assignment's own config map, which has
no `NVR_MQTT_PASSWORD`, and fails with `required config 'NVR_MQTT_PASSWORD'
has no value`. **Assigning nvr to any node now fails.**

Confirmed on node 1314, which is recording 3 cameras today: its
`~/citadel-node/services/nvr.env` has 9 keys and no `NVR_MQTT_PASSWORD`, so the
merged module cannot be redeployed there at all.

## Root cause

`ConfigVar` offers exactly three ways to get a value: an override, a static
`default:`, or an interactive prompt. A credential used only BETWEEN a module's
own containers fits none of them — no operator has a value to type, no upstream
has one to send, and a `default:` would ship one shared password to every node
in the fleet.

## Changes

- `internal/catalog/catalog.go`: `ConfigVar.Generate` (`generate: secret`) — the
  node mints a random 256-bit URL-safe token for itself. URL-safe and unpadded
  so the value carries no `=`, quote, or shell metacharacter into compose env
  and quoted entrypoints.
- `internal/catalog/install.go`: generation slots between the override and
  default checks, so an explicit value still wins (that is how you point the
  module at an external broker). Generated values are **sticky**: install reads
  the module's existing `.env` first and reuses what is there. An unknown
  `generate:` kind is a loud error — silently returning `""` would hand a broker
  an empty password.
- `internal/catalog/install.go` + `cmd/module_ops.go`: `CarryGeneratedConfig`.
  The assignment path updates in place by uninstalling first, and uninstall
  deletes `<name>.env` — so read-back alone is not enough there and the secret
  would rotate on every re-assignment. The persisted value is now carried across
  the teardown. See "why rotation is not benign" below.
- `services/nvr-service/service.yaml`: `NVR_MQTT_PASSWORD` gains
  `generate: secret`, keeping `required: true` so the compose `:?` still fails
  closed.

Generic, not nvr-specific: any module with an internal credential can use it.

## Why rotation is not benign

Compose recreates only the containers whose env changed. `NVR_MQTT_PASSWORD` is
env for `mosquitto` and `nvr-config`, but **not** for `frigate` — Frigate reads
the regenerated `config.yml`. So a rotation recreates the broker and leaves
Frigate running with the old credential in memory; it goes quiet until it is
restarted explicitly. Observed for real on node 1314 while rotating by hand.
Sticky reuse is what keeps ordinary reconciles from tripping this.

## Scope — what this does NOT fix

`cmd/module_update.go:169` (`citadel module update`) passes `nil` overrides
non-interactively, and `existing` is consulted only for `generate:` vars. So
`WYZE_EMAIL` / `WYZE_PASSWORD` / `API_ID` / `API_KEY` / `NVR_CAMERAS` — required,
no default, no generate — still fail there. `citadel module update nvr` remains
broken. That is pre-existing (it predates #639) and out of scope here.

Nothing in this chain is released: `v2.96.0` predates #638, #639, #644 and this
PR. A release is needed before any node other than 1314 (running a dev build)
can deploy nvr.

## Testing

`go test ./internal/catalog/ ./services/ ./cmd/` — new coverage for:
required-is-satisfied non-interactively (the exact regression), per-node
uniqueness, stickiness across installs, stickiness through a real `.env` round
trip, carry-across-teardown (including that non-generated vars are NOT revived
from a stale `.env`, and that the caller's config map is not mutated — it is
what the lockfile records for drift detection), override precedence,
unknown-kind error, and `.env` parse tolerance.

Pre-existing macOS-local failures in `internal/jobs` / semantic search are
unrelated and reproduce on a clean `main` worktree.

Deploy verification on node 1314 in a PR comment below.

### DIFF
--- a/cmd/module_ops.go
+++ b/cmd/module_ops.go
@@ -151,16 +151,25 @@ func (o *liveModuleOps) Install(ctx context.Context, m reconcile.ModuleAssignmen
 	// Residual (interim, acceptable): if this Uninstall succeeds but the
 	// InstallFromManifest below then fails, the module is left down until the
 	// job retries and reinstalls it.
+	// Node-generated secrets must survive the teardown below: Uninstall deletes
+	// <name>.env, so without carrying them forward here the re-install would mint
+	// a NEW value on every re-assignment -- and compose would then recreate only
+	// the container whose env changed, leaving its consumer running with the old
+	// credential in memory. Read BEFORE the uninstall; anything the assignment
+	// supplies still wins, so an explicit rotation is still possible.
+	installConfig := m.Config
 	if hasService(nodeManifest, manifest.Name) {
+		installConfig = catalog.CarryGeneratedConfig(manifest, servicesDir, m.Config)
 		o.log("MODULE_SET: %q already installed; updating in place", manifest.Name)
 		if err := o.Uninstall(ctx, manifest.Name); err != nil {
 			return fmt.Errorf("update %q: uninstall existing: %w", manifest.Name, err)
 		}
 	}
 
-	// Non-interactive install: m.Config supplies the overrides; a missing REQUIRED
-	// config var is a returned error (never a stdin prompt on a headless node).
-	result, err := catalog.InstallFromManifest(manifest, composeSrc, servicesDir, m.Config, false, allowPrivileged, untrusted, false)
+	// Non-interactive install: installConfig supplies the overrides; a missing
+	// REQUIRED config var is a returned error (never a stdin prompt on a headless
+	// node).
+	result, err := catalog.InstallFromManifest(manifest, composeSrc, servicesDir, installConfig, false, allowPrivileged, untrusted, false)
 	if err != nil {
 		return fmt.Errorf("install %q: %w", manifest.Name, err)
 	}
--- a/internal/catalog/catalog.go
+++ b/internal/catalog/catalog.go
@@ -236,7 +236,27 @@ type ConfigVar struct {
 	Description string `yaml:"description"`
 	Default     string `yaml:"default"`
 	Required    bool   `yaml:"required"`
+	// Generate names a value the NODE mints for itself when the operator supplies
+	// none, instead of failing or prompting. The only kind is "secret": a random
+	// 256-bit URL-safe token.
+	//
+	// This exists for credentials that exist purely BETWEEN a module's own
+	// containers -- a broker password, an internal API key -- which nobody
+	// outside the node ever needs to know or type. Without it such a var has no
+	// good option: `required: true` hard-fails the non-interactive assignment and
+	// update paths (a fabric MODULE_SET carries no value for it), while a static
+	// `default:` would ship the same well-known credential to every node.
+	//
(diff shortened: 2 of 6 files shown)


sunapi386:
Deploy-verified on node 1314 (mini), which records 3 cameras to the Synology NFS export.

## The regression, reproduced

Both binaries were run through the same install path with every required
config var supplied as an override EXCEPT `NVR_MQTT_PASSWORD` — i.e. exactly
what a fabric `MODULE_SET` assignment carries — with stdin closed so a prompt
counts as a failure:

```
=== OLD binary (pre-fix) ===
Installing nvr from catalog...
  NVR_MQTT_PASSWORD (...): Error: install failed:
  aborted: no value provided for required config 'NVR_MQTT_PASSWORD'
OLD rc=1

=== NEW binary (generate: secret) ===
Installing nvr from catalog...
Installed nvr successfully.
NEW rc=0
```

The minted secret is 43 chars (32 random bytes, base64url unpadded) and was
persisted to `services/nvr.env`.

## Stack verification

```
/dev/shm            512M  (was 256M; Frigate wanted >= 306M)
shm warning         gone from the Frigate startup log
config.yml mqtt     enabled: true, host: mosquitto, port: 1883, topic_prefix: frigate
containers          citadel-nvr-{frigate,mosquitto,wyze-bridge} all Up, frigate healthy
cameras             garag3, goulart3, lab — recording again
```

Broker auth, asserted on exit code and payload (not just "container up"):

```
authed subscribe    rc=0   -> frigate/garag3/enabled/state ON, .../recordings/state ON, ...
anonymous           rc=5   -> Connection Refused: not authorised
wrong password      rc=5   -> Connection Refused: not authorised
published ports     none   -> reachable only on the module compose network
```

`citadel run nvr` also exercised #638: podman is preferred on this node but
its socket is dead, and the runtime correctly fell back to docker with the
warning.

## One operational note worth keeping

The generated secret is deliberately sticky, and this run showed why. When I
rotated it by hand, compose recreated `mosquitto` (its env changed) but NOT
`frigate` (which takes no MQTT env — it reads the regenerated `config.yml`).
Frigate kept the old credential in memory and went quiet until it was
restarted explicitly. So a *rotation* is a two-container operation; the sticky
reuse in this PR is what keeps ordinary reconciles from silently triggering
that.

## Unrelated finding

Restarting the worker to install this build dropped the node's gateway
exposure — `/expose/frigate` went 404 while the gateway root stayed 200, and
re-running `citadel service expose` restored it. Exposures are in-memory only.
Filed as #647; not a regression from this PR.

Issue #647, opened 2026-07-31
Gateway exposures are lost on every worker restart (in-memory only)


sunapi386:
Found while deploying #646 to node 1314 (mini).

## What happens

Exposures created via `citadel service expose` / the `/agent/expose` control
endpoint / the MCP `expose` verb live only in the running `citadel work`
process. Restart the worker and every exposure silently disappears.

Reproduced on node 1314:

```
# before restarting the worker
GET https://100.64.0.97:8443/expose/frigate/   -> 200

# after `systemctl restart citadel-worker`
GET https://100.64.0.97:8443/expose/frigate/   -> 404
GET https://100.64.0.97:8443/                  -> 200   # gateway itself is fine

# re-running `citadel service expose frigate --port 8212 --visibility org`
GET https://100.64.0.97:8443/expose/frigate/   -> 200
```

So the gateway is healthy — only the exposure set is gone. `~/.citadel-cli/`
holds an `expose_signing_key` (so link tokens survive) but no exposure state
file, which is the tell: the signing key was made durable, the routes it signs
for were not.

## Why it matters more than it looks

A worker restart is NOT a rare event. It happens on auto-update, on
`WORKER_SELF_HEAL` (#548), on any `systemctl restart`, and on every reboot.
So an exposure the user set up from the iOS app (aceteam#6799) or the MCP
`expose` verb (aceteam#6650) will stop working at an arbitrary later time,
with no notification and no error — the link just starts 404ing.

That is the whole "share a service on the fabric" product surface silently
decaying. It also makes an exposure impossible to reason about from the
platform side, since nothing off-node knows the set was dropped.

## Suggested fix

Persist the exposure set next to the signing key (e.g.
`~/.citadel-cli/exposures.json`, 0600) and re-wire it at gateway start, the
way `gateway-facts.json` already persists gateway state. Points to settle:

- **Expiry:** a `link` exposure carries a TTL. Restoring an expired one must
  be a no-op, and a restored token must still validate against the persisted
  epoch so revocation-by-epoch-bump keeps working.
- **Port liveness:** the upstream port may be gone after a reboot (module
  removed, port changed). Restoring a route to a dead port should be visible,
  not a silent 502.
- **Ordering:** restore before the gateway starts serving, so there is no
  window where a valid exposure 404s.

## Scope note

This is orthogonal to the `private`-collapses-to-org identity gap (every node
joins Headscale as `org_{uuid}`, so `private` authorizes any org peer rather
[...]

Issue #548, opened 2026-07-17
bug(worker): job consumption wedged for 4+ hours while heartbeats kept flowing (node 1084, v2.78.0)


sunapi386:
## Observed (node 1084, 2026-07-16/17)
- Last consumed job: 21:07 local (transcribe). From then until 01:25 UTC the worker consumed NOTHING: a trivial `dispatch_fabric_job` echo timed out, and two /fabric/models/deploy dispatches sat unread in the per-node stream.
- Heartbeats continued the whole time (fabric_node_status.reported_at stayed current; UI showed node online) — so the wedge is in the job-consumption path, not the process.
- `citadel service stop && start` fixed it instantly; the new worker (PID 706599) drained the entire backlog (3 MODEL_CACHE_PULL + 3 SERVICE_START + echo) within seconds of starting.

## Context before the wedge
The last activity before silence was a meeting/transcribe loop repeatedly failing with `open .../meetings/<id>-window.wav: permission denied` and `Waiting for transcription service to become ready...` (21:04–21:07), plus `max meeting duration (5m0s) reached; leaving` at 21:06:48. Suspect: a job handler (transcribe/meeting?) blocking the consumer loop indefinitely — no per-job timeout on the consumption path?

## Impact
Node looks healthy (green, heartbeating) but silently executes nothing — deploys, shell jobs, everything dead-letters into wait. This defeats the fail-closed routing guarantees (#4500): resolve_shell_queue sees the per-node subscription as live (consumer group exists) and dispatches into a void.

## Suggested investigation
- Per-job execution timeout / watchdog in the worker consumer loop (if a handler hangs, log + move on or restart consumer).
- A liveness probe distinct from heartbeat: publish 'last job consumed at' in the heartbeat so the platform can flag consumption-wedged nodes.
- Reproduce: inspect what the transcribe/meeting handler does when the WAV is permission-denied in a retry loop.

sunapi386:
## Folding in: no MCP tool actually restarts a wedged worker

Confirmed while recovering node 1084 (worker healthy again after a systemd restart). Reviewed the recovery tooling and there is a real gap on the remote-control side:

| MCP tool | Reality |
|----------|---------|
| `citadel_worker_status` | ✅ inspects (consuming?, per-node subscription, last_poll/last_job, consume HTTP status) |
| `citadel_doctor` | ✅ diagnoses 'why no per-node jobs' |
[...]

PR #648, opened 2026-07-31
fix(gateway): persist exposures so they survive a worker restart (#647)

sunapi386:
## Context

Exposures created via `citadel service expose`, the `/agent/expose` control
endpoint, or the MCP `expose` verb lived ONLY in the running `citadel work`
process. Restarting the worker silently dropped every one of them.

Found while deploying #646 to node 1314. Reproduced there:

```
GET https://100.64.0.97:8443/expose/frigate/   -> 200
# systemctl restart citadel-worker
GET https://100.64.0.97:8443/expose/frigate/   -> 404
GET https://100.64.0.97:8443/                  -> 200   # gateway itself fine
# citadel service expose frigate --port 8212 --visibility org
GET https://100.64.0.97:8443/expose/frigate/   -> 200
```

Worker restarts are routine — auto-update, WORKER_SELF_HEAL (#548),
`systemctl restart`, reboot — so a link shared from the iOS app (aceteam#6799)
or the MCP verb (aceteam#6650) stops working at an arbitrary later time, with
no error and nothing off-node aware the route was dropped.

## Root cause

The link-token SIGNING KEY was already made durable (`expose_key.go`) so a
token minted yesterday still verifies today — but the routes those tokens
authorize were not. The token outlived the thing it opened.

## Changes

- `internal/config/exposures.go`: `ExposureRecord` + Load/Save/Delete, stored
  0600 as `exposures.json` beside the signing key, written temp-file+rename so
  a crash mid-write cannot truncate the store and drop every exposure. The
  link TOKEN is deliberately NOT stored — tokens are derived from the key and
  carry their own expiry; persisting one would outlive its minter.
- `cmd/expose_ops.go`: `liveExposeOps.Expose` persists after programming the
  gateway. Every caller (CLI, MCP verb, EXPOSE_SET job) funnels through here,
  so live and durable sets cannot diverge. A write failure is LOGGED, not
  returned: the exposure is live and the caller already has its URL, so
  failing would be a lie — the honest failure is "works now, gone after a
  restart", and it must be visible.
- `cmd/expose_ops.go`: `restoreExposures` re-wires the set, called from
  `work.go` and `serve.go` BEFORE the gateway serves — restoring afterwards
  would leave a window in which a valid exposure 404s, the exact symptom being
  fixed. A record the gateway rejects is skipped loudly rather than dropping
  the whole set, and a restored route whose upstream is not listening logs a
  warning (otherwise it restores fine and every request 502s with nothing
  explaining why). All failures are non-fatal: a node with an unreadable
  exposure store must still come up serving its builtin routes.

`TokenEpoch` is preserved explicitly — resetting it to 0 would break every
live link token and make a later epoch bump unable to revoke what it had
already handed out.

## Testing

`go test ./internal/config/ ./cmd/ ./internal/gateway/` — round trip,
epoch preservation, replace-by-name (a duplicate would let restore order pick
the policy and could reinstate a looser visibility), missing store is not an
error, corrupt store IS an error but does not wedge future saves, delete is
idempotent, 0600 permissions, no stray temp file.

On-node verification on 1314 to follow in a PR comment.

## Known gap (unchanged by this PR)

There is still no unexpose path — `Server.RemoveExposure` has no caller
outside tests. `DeleteExposure` is here so that when one lands it has a
durable counterpart.

### DIFF
--- a/cmd/expose_ops.go
+++ b/cmd/expose_ops.go
@@ -12,6 +12,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"net"
 	"time"
 
 	"github.com/aceteam-ai/citadel-cli/internal/config"
@@ -49,6 +50,22 @@ func (liveExposeOps) Expose(_ context.Context, req worker.ExposeRequest) (*worke
 		return nil, err
 	}
 
+	// Persist so the exposure survives a worker restart (#647). Every caller --
+	// the CLI, the MCP verb, the EXPOSE_SET job -- funnels through here, so this
+	// is the one place the durable set can be kept in step with the live one. A
+	// write failure is logged, not returned: the exposure IS live and the caller
+	// already has its URL, so failing the call would be a lie; the honest failure
+	// mode is "works now, gone after a restart", and it must be visible.
+	if err := config.SaveExposure(platform.ConfigDir(), config.ExposureRecord{
+		Name:       req.Name,
+		Port:       req.Port,
+		Visibility: req.Visibility,
+		Creator:    req.Creator,
+		TokenEpoch: req.Epoch,
+	}); err != nil {
+		Log("warning: exposure %q is live but was not persisted (it will not survive a restart): %v", req.Name, err)
+	}
+
 	res := &worker.ExposeResult{URL: exposeMeshURL(req.Name)}
 
 	if policy.Visibility == gateway.VisibilityLink {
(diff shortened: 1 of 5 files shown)


sunapi386:
On-node verified on 1314 (mini).

```
=== before ===
~/.citadel-cli/exposures.json                        does not exist

=== create ===
citadel service expose frigate --port 8212 --visibility org
  URL: https://100.64.0.97:8443/expose/frigate

-rw------- 1 jason jason 76 ~/.citadel-cli/exposures.json
[ { "name": "frigate", "port": 8212, "visibility": "org" } ]

=== systemctl restart citadel-worker  (the action that used to drop it) ===
worker log: "Restored 1 gateway exposure(s)"

GET /expose/frigate/  from an org peer   -> 200
  and the asset rewrite survived: src="/expose/frigate/assets/main-CosNs8ey.js"
```

Same request was **404** after a restart before this change.

Access control is unchanged and still fails closed on the restored route — it is a real re-registration, not a bypass:

```
self (not an org peer)     -> 403
unknown exposure name      -> 404
org peer                   -> 200
```

File is 0600 and, as intended, carries no token — only name/port/visibility (plus creator/epoch when set).

Issue #643, opened 2026-07-30
Epic: `citadel up` — machine-wide TUN mode, pure Go, no Tailscale install


sunapi386:
## Goal

A user should never need to install Tailscale.app to put their whole machine on the fabric. `citadel up` brings up a real TUN interface so **every** app on the box can reach `100.64.x.x` — `ssh`, `curl`, a browser, a database client — with citadel as the only binary involved.

Explicit constraints from Jason (2026-07-30):

1. **Pure Go dependencies only.** No bundled C libraries, no Tailscale.app, no separate driver install we ship.
2. **Root/admin is opt-in.** `citadel login` (userspace `tsnet`) stays the unprivileged default and keeps working exactly as it does now. `citadel up` is the thing a user *chooses* when they want machine-wide routing, and only that path may ask for elevation.

## Feasibility: verified, not assumed

The dependencies are already in the module graph — `tailscale.com v1.100.0` is a direct dependency, so `tstun` and `wgengine` cost us nothing new. Confirmed they build **CGO-free** on every target:

```
CGO_ENABLED=0 go build tailscale.com/net/tstun tailscale.com/wgengine
darwin/arm64:    OK — builds pure Go
linux/amd64:     OK — builds pure Go
windows/amd64:   OK — builds pure Go
```

So constraint (1) is satisfiable. The work is wiring, privilege handling, and lifecycle — not a dependency hunt.

## Architecture

Today `internal/network/server.go` constructs a `tsnet.Server`, which is userspace netstack: it gives the *process* a mesh identity via `srv.Dial` / `srv.Listen`. TUN mode instead creates a real interface via `tstun.New()` and hands it to `wgengine`, so the kernel routes for the whole machine.

Suggested shape: a `NetworkServer` backend interface with two implementations (`userspace` = today's tsnet, `tun` = new), so every existing caller of `Dial`/`Listen` is backend-agnostic and `citadel login` is untouched.

## Per-platform reality

| Platform | Device | Privilege | Pure Go? |
|---|---|---|---|
| macOS | `utun` via `tstun.New()` | root (`sudo citadel up`) | Yes |
| Linux | `/dev/net/tun` | root, or `CAP_NET_ADMIN` on the binary | Yes |
| Windows | Wintun | Administrator | Yes, **but see below** |

[...]

PR #650, opened 2026-07-31
feat(network): machine-wide mode (`citadel up`) — Backend split, TUN + attached backends (#643)

sunapi386:
Implements #643 through slice 3, plus the Windows adapter work. Supersedes this PR's original scope (it opened as the interface extraction alone — see the commit-by-commit breakdown below).

## What this does

`citadel login` connects the citadel **process**. `citadel up` puts the **whole machine** on the mesh via a real kernel TUN, so `ssh`, `curl`, a browser — anything on the box — reaches `100.64.x.x` directly. Root/admin only, strictly opt-in, and it never silently downgrades.

The design decision Jason made on 2026-07-31 was **option 3 from `docs/machine-wide-tun.md`: one machine is one node.** `citadel up` shares the state dir — and therefore the node identity — with `citadel login`, and every other citadel process on that host **attaches** to it instead of starting a second WireGuard endpoint. Callers needed no changes: they already go through `network.Dial` / `network.Listen` / `network.GetGlobalPeers`.

## The load-bearing finding

`tsnet.Server` has a public `Tun tun.Device` field that looks like it would give machine-wide routing. It does not. tsnet builds `wgengine.Config{Tun: s.Tun, ...}` and sets **neither `Router` nor `DNS`**, so wgengine substitutes `router.NewFake()` and a no-op DNS configurator — packets would cross a real interface while the OS routing table went untouched and the resolver unconfigured. There is no hook to supply them, so `backend_tun.go` assembles the tailscaled-style stack itself.

Two requirements in that stack are easy to miss and break the node quietly:
- **`netns.SetEnabled(true)`** — binds outbound WireGuard packets to the physical interface so they do not re-enter the tunnel. tsnet never needs it (always netstack, so no tunnel to loop through).
- **`dns.CleanUp` + `router.CleanUp` on every `Up`**, not just teardown. A crash or reboot-without-shutdown strands routes and a rewritten resolver that restarting would not otherwise undo. tailscaled does the same, for the same reason.

## Commits

| | |
|---|---|
| `93409c3` | Backend interface; today's tsnet becomes `backend_userspace.go`. No behavior change. |
| `53b5d33` | Pin the authkey through the backend constructor (see "Verification" — the live check could not reach it). |
| `2033179` | `tunBackend`, `attachedBackend`, `SelectBackend`, `citadel up`/`down`. |
| `d657303` | Windows adapter identity + `citadel up --check`. |
| `fa52675` | Three crashes in the TUN path that `--check` could not reach. |

## Windows: citadel needed its own Wintun identity

`tstun`'s init pins **both** the tunnel type (`"Tailscale"`) and a **static adapter GUID**. Wintun keys on the GUID, so every tstun consumer asks for the same adapter and the interface name is irrelevant. On the Windows 11 test VM (which runs Tailscale), the first `citadel up --check` failed with `0x800700B7` — "file already exists". `internal/network/tun_windows.go` claims citadel's own type and GUID; its init runs after tstun's, so it overrides rather than races. The GUID is fixed, not random, so restarts re-attach instead of leaking an adapter per run.

## Verification — please read this part

**Verified live:**
- The refactored **userspace** path against the real mesh on macOS: ONLINE at 100.64.0.21, correct per-peer direct/relay classification. This covers `Connect` → `Up` → `waitForConnection` → `Status` → `GetPeers` → `PingPeer`.
- **Windows**, as SYSTEM on DESKTOP-6UKHJAN: binary runs; elevation detected; `citadel up --check` creates and removes citadel's Wintun adapter; the adapter is gone afterwards and Tailscale is undisturbed, still on the mesh at 100.64.0.110.
- Unelevated `citadel up --check` on macOS reports the elevation requirement and exits non-zero rather than falling back.

**NOT verified — no live `citadel up` has ever been run:**
- The full TUN bring-up: engine start, route installation, MagicDNS, teardown.
- That `citadel up` and `citadel login` really resolve to **one** node. Commit `fa52675` fixes the flag that would have silently split them (`LocalBackendStartKeyOSNeutral`), but only a bring-up can confirm the node count does not grow.
- Route-level coexistence with an installed Tailscale. `--check` installs no routes; both products want `100.64.0.0/10`. The docs now say this explicitly — an earlier draft overclaimed it.
- Linux (`/dev/net/tun`) end to end.

I did not do the first bring-up on Jason's Mac: a failed teardown there breaks DNS on the working machine. It wants a disposable Linux host.

**A note on how the last commit was found.** `--check` skips `tunBackend.Up` and `CleanUpSystemState`, so that code had zero runtime coverage. Reading the tailscale module turned up three bugs that would all have fired on the first real `citadel up`: `sys.Set(dev)` hitting a `default: panic`, `loginFlags=0` silently splitting the node identity, and `CleanUpSystemState` dereferencing a nil event bus — the last taking out `citadel down`, the recovery path, along with it.

## Deliberate non-changes

- **Selection does not refuse concurrent userspace processes.** A background `citadel work` plus an ad-hoc `citadel status` — two tsnet endpoints on one state dir — is long-standing behavior. An earlier draft made it an error, which would have broken every node for a problem this PR did not introduce. Only the new userspace/TUN collision is blocked, in `ConnectMachineWide`, naming the blocking pid.
- **`citadel up` stays hidden from the curated top-level help**, matching `login` and the "just run `citadel`" philosophy in `hideAdvancedCommands`. Say the word if it should be promoted.

## Tests

25 cases in `internal/network`. New here: selection order, the userspace-does-not-block-userspace guarantee, dead-pid tolerance, live-holder errors naming their pid, refusal to delete a non-socket, a stale socket not being mistaken for a live backend, and `CleanUpSystemState` not panicking. Compile-time assertions pin all three backends to the interface.

`go build` / `go vet` / `go test` clean. `internal/network` cross-compiles CGO-free for linux/amd64, linux/arm64, windows/amd64, darwin/arm64. (A full `go build ./...` for windows fails in `internal/nvr` on `syscall.Stat_t` — pre-existing on `main`, unrelated, not on the release path.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sunapi386:
## Live verification done — promoting out of draft

The manual check the PR body said was still owed has now been run. Built from this branch (`go build ./cmd/citadel`) and pointed at the real mesh on macOS (jmbp16-m1pro, node 100.64.0.21):

```
🌐 NETWORK STATUS
  Connection:  🟢 ONLINE to AceTeam Network
  Hostname:    jmbp16-m1pro
  IP Address:  100.64.0.21
  Peers:
    🟢 mini 100.64.0.97 91ms [direct] (linux)
    🟢 glassbox 100.64.0.33 193ms [relay:ace] (windows)
    🟢 ubuntu-gpu 100.64.0.78 5ms [direct] (linux)
```

That single command exercises the whole refactored path end to end: `Connect` → `backend.Up` → `waitForConnection` (polling `Backend.Status`) → `Status` summary → `GetPeers` same-user filter → `PingPeer`. The IP is unchanged, and the direct/relay classification still comes out right per peer — `glassbox` over DERP `ace`, the other two direct.

Ready for review.

sunapi386:
## Correction to the comment above, plus the gap it was hiding

My previous comment overstated what was verified. The PR body said the owed manual check was **`citadel login` on a real node**. What I actually ran was **`citadel status`** — and those are not the same path in the way that matters here.

`citadel status` reaches the mesh through `VerifyOrReconnect` / `EnsureConnected`, and **both call `Connect(ctx, "")`**. So the run exercised `Connect` → `Up` → `waitForConnection` → `Status` → `GetPeers` → `PingPeer` against the live coordination server — real and useful — but it never touched the one field this refactor moved:

```go
newUserspaceBackend(ServerConfig{...}, s.stateDir, authKey)  // → tsnet.Server.AuthKey
```

Had that plumbing been dropped, a node reconnecting from saved state would still have gone green while `citadel login`, `citadel init --authkey` and `ReconnectWithAuthKey` silently failed to authenticate. The fake-backend tests can't catch it either, because `Connect` constructs the backend internally.

**Closed in 53b5d33** with two construction tests asserting `AuthKey` / `Dir` / `Hostname` / `ControlURL` reach `tsnet.Server`, and that `Ephemeral` stays false. Confirmed as a real guard rather than a tautology — stubbing `AuthKey` to `""` fails the test:

```
--- FAIL: TestUserspaceBackendCarriesConfig
    backend_test.go:91: AuthKey = "", want hskey-abc
```

That commit also cross-compiles the package, since `release.sh` builds all platforms: `internal/network` builds CGO-free for windows/amd64, linux/amd64, linux/arm64 and darwin/amd64, as does the `./cmd/citadel` release target. (A full `go build ./...` for windows fails in `internal/nvr` on `syscall.Stat_t` — pre-existing on `main`, unrelated to this PR, and not on the release path.)

**So the accurate verification status is:** the reconnect-from-saved-state path is verified live on the mesh; the authkey path is covered by construction tests, not by a live `citadel login`. A reviewer who wants belt-and-braces should run `citadel login` on a node before merging.

sunapi386:
## First real `citadel up` — on a genuinely clean Windows machine

Ran on VM 109 (Windows 11, **no Tailscale, no `wintun.dll`**) — the actual "user who never installed Tailscale" scenario, which is the whole point of #643. It found **two runtime-only bugs** that compile fine on every platform, and one of them meant machine-wide mode was broken *everywhere*, not just on Windows.

### 1. `create router: unsupported OS "windows"` — affected all platforms

`router.New` dispatches through a feature hook (`router.HookNewUserspaceRouter`) that is only populated by importing the implementation package. The TUN device was created, then the bring-up died. The same hook gates Linux and macOS, so this was not a Windows problem — machine-wide mode simply had no router anywhere.

tailscaled imports `_ "tailscale.com/feature/condregister"`. I started there and backed it out: that umbrella also registers Taildrive, TPM, and the Kubernetes and AWS state stores. **Measured: it links the AWS SSM client and 76 aws/smithy packages into citadel.** Importing just `_ "tailscale.com/wgengine/router/osrouter"` gives the routers with **zero** AWS packages linked. The four indirect deps it does add (netlink, nftables, go-iptables, netns) are the Linux routing libraries the router genuinely needs.

### 2. `CoInitialize has not been called` ×20, then "giving up"

osrouter's `setPrivateNetwork` — which marks citadel's adapter **Private** so Windows Firewall does not treat mesh peers as untrusted — builds an `ole.Connection` carrying the comment *"DO NOT call Initialize() ... We've already handled that process-wide"*. tailscaled does that in its own init; nothing in the library does it for you.

This one **does not crash**. The engine starts anyway, so the only symptom is an adapter silently left in the **Public** firewall profile — a security-relevant misconfiguration that reports success.

### Results

```
citadel up --check, no wintun.dll present:
  Network device: FAILED — Error loading wintun.dll DLL

with wintun.dll beside the binary (installs the driver from scratch):
  Installing driver 0.14 / Extracting driver / Creating adapter
  Network device:  yes ({C17ADE10-...})

citadel up:
  Name       Status
  Ethernet 2 Up
  citadel0   Up        <- real TUN interface, no Tailscale installed
  CoInitialize errors: 0
```

**Teardown is clean on the failure path** — the property I most wanted to check. After the process exited: adapter removed, **zero** `100.x` routes left, no orphaned process.

### Still unverified

Auth times out (this node has no authkey and has never joined), so **route installation and MagicDNS are still untested**, and so is the one-machine-one-node claim. Getting past that needs an authkey for a throwaway node — happy to run it if you generate one, or it can wait for the Linux bring-up.

### Bonus finding for slice 6 (the `wintun.dll` embed)

The loader is not a general DLL search:

```go
windows.LoadLibraryEx(d.Name, 0,
    LOAD_LIBRARY_SEARCH_APPLICATION_DIR|LOAD_LIBRARY_SEARCH_SYSTEM32)
```

Those two flags are the *entire* search path — not `PATH`, not the working directory, not any path we choose. So "extract to an Administrator-only-writable directory" is only satisfiable by the binary's own directory or System32. That suits the security requirement rather than fighting it, but it makes the open question a **permissions** one (verify the application directory is not attacker-writable before loading — an unprivileged `%LOCALAPPDATA%` install is writable by the same user who runs `citadel up` elevated) rather than a packaging one.

License checked: Wintun's §3d permits redistribution "insofar as the Software is distributed alongside other software that uses the Software only via the Permitted API" — exactly citadel's use.

VM 109 is shut down and cleaned; no adapter, routes, or files left behind.

sunapi386:
## End-to-end bring-up: **works**. The core claim is now proven.

You were right that I could mint the authkey myself — `nexus_generate_authkey` was available and I asked you for something I could do. (The MCP tools then hung for 25+ minutes, so I minted it directly on ocean via `headscale preauthkeys create`; separate issue, noted below.)

Two more bugs stood between the TUN stack and a registered node.

### 1. A fresh node never registered

`lb.Start` does **not** begin login on its own. It calls `cc.Login` only when a node key already exists or a config file set WantRunning:

```go
if !loggedOut && (b.hasNodeKeyLocked() || confWantRunning) {
        cc.Login(b.loginFlags)
}
```

A machine that has never joined satisfies neither, so it parked in `NeedsLogin` with `authRoutine: loggedIn=false; goal=nil` — the authkey sat on the control client and nothing ever asked it to log in. `Up()` then failed with **"timeout waiting for network connection"**, which reads like a network problem and is not one. tsnet hits the same gate and handles it the same way (tsnet.go:960).

### 2. `citadel up` never wired up logging

It never called `network.SetLogf`, so the package-level `logf` stayed a no-op and **every engine diagnostic was discarded**. Bug 1 was one opaque line. With logging wired it took a single run to spot `Switching ipn state NoState -> NeedsLogin` / `goal=nil`. `citadel work` and `citadel reconnect` already did this; `up` was the odd one out.

### Results — Windows VM 109, clean machine, no Tailscale

```
headscale:  tun-testvm109 | 100.64.0.46 | online
interface:  citadel0 Up, 100.64.0.46
routes:     full peer table + 100.100.100.100/32 (MagicDNS)
firewall:   citadel0 = Private          <- the COM fix, confirmed working
```

**Machine-wide routing, from tools that know nothing about citadel** — this is the thing tsnet could never do:

```
ping 100.64.0.78                     ->  3/3 replies, avg 3ms
Resolve-DnsName ubuntu-gpu.internal  ->  100.64.0.78
```

**One machine, one node** — the load-bearing claim of the design:

| | org node count |
|---|---|
| baseline | 12 |
| after `citadel up` | 13 (+1) |
| after `citadel status` alongside it | 13 (**+0**) |

`citadel status` reported the same node and IP as the running `citadel up` — it attached rather than starting a second WireGuard endpoint. Exactly one `tun-testvm109` row throughout.

**Teardown after `Stop-Process -Force`** (no graceful shutdown — the crash path): adapter removed, **0** leftover `100.x` routes, DNS restored to the LAN resolver.

### Status

Everything in the "NOT verified" list from the PR body is now verified, except **route-level coexistence with an installed Tailscale** (VM 109 has none, by design — it is the clean-machine test). That is the last gap, plus a Linux bring-up.

Cleaned up: test node deleted from headscale (back to 12), preauth key expired, `C:\citadel-test` removed, VM shut down.

### Unrelated: the aceteam MCP nexus tools appear to hang

`nexus_list_nodes` and `nexus_generate_authkey` both ran **25+ minutes without returning** and had to be killed. `nexus.aceteam.ai/health` and `aceteam.ai/api/health` were both 200 in <200ms at the time, and the same operations via `headscale` CLI on ocean returned instantly. Worth a look — I did not investigate further since it is outside this PR.

sunapi386:
## All three platforms verified — out of draft

#676 (macOS MagicDNS) is fixed and closed, which was the last functional gap. Coverage now:

| | Windows | Linux | macOS |
|---|---|---|---|
| TUN interface + routes | ✓ | ✓ | ✓ |
| Reach peer by IP, from a non-citadel tool | ✓ | ✓ | ✓ |
| MagicDNS name resolution | ✓ | ✓ | ✓ (needed #676) |
| One machine, one node | ✓ | ✓ | ✓ |
| Clean teardown | ✓ (incl. SIGKILL) | ✓ | ✓ |

Test hosts were chosen to be adversarial rather than convenient:

- **Windows** — a genuinely clean VM: no Tailscale, no `wintun.dll`. The "user who never installed Tailscale" case. Driver installed from scratch.
- **Linux** — `ubuntu-gpu`, a **live fabric node** already running `citadel work` under user systemd **and** `tailscaled`. It reused the node's existing identity (came up as its own 100.64.0.78, org node count unchanged at 12), and `tailscale0` kept its address throughout and survived citadel's teardown untouched. That closes route-level coexistence.
- **macOS** — the working machine, with Tailscale.app removed so citadel was the only thing managing the tunnel. Reused the existing `citadel login` identity with no authkey.

## What the live testing actually bought

Six bugs that unit tests and `--check` could not reach, three of them platform-independent:

1. `sys.Set(dev)` with a raw `tun.Device` hits `tsd.System.Set`'s `default: panic`.
2. `loginFlags=0` instead of `LocalBackendStartKeyOSNeutral` would have **silently split the node identity** — the exact failure this design exists to prevent, while printing success.
3. `CleanUpSystemState` dereferenced a nil event bus, taking out `citadel down` — the recovery path — along with startup.
4. `router.New` returned `unsupported OS` on **every** platform: the implementation lives behind a feature hook only populated by a blank import.
5. Windows needed process-wide COM init, or the adapter is silently left in the **Public** firewall profile.
6. macOS dropped the MagicDNS nameservers entirely (#676).

Most of these fail *quietly* — success messages, healthy-looking logs, working routing. That is what makes `citadel up` worth the live verification rather than a green test suite.

## Reviewer attention

- **`internal/network/dns_darwin.go` changes macOS from global DNS to split DNS.** Deliberate, and a departure from what upstream's manager intends. Rationale in the file and on #676.
- **`SelectBackend` does not refuse concurrent userspace processes.** An earlier draft did, which would have broken `citadel status` on every node running `citadel work`. Only the new userspace/TUN collision is blocked.
- **Two blank imports are load-bearing and invisible to the compiler** — `osrouter`, and the Windows COM init. Both are commented; removing either breaks machine-wide mode at runtime only.

## Known gaps, not blocking

- Reverse DNS on macOS (~65 `.arpa` domains would mean ~65 `/etc/resolver` files). Forward names work.
- The `wintun.dll` embed is still outstanding — Windows currently needs the DLL beside the binary. The loader constraint is documented in `docs/machine-wide-tun.md`.
- Follow-ups filed: #662 (`SetLogf` unwired — the bug that cost the most time here), #663 (`internal/nvr` Windows cross-compile), #670 (Citadel.app), #672 (macOS signing).

PR #661, opened 2026-08-01
feat(gateway): add unexpose to revoke a gateway exposure

sunapi386:
## Context

The last gap from #647. Exposures could be created (`citadel service expose`,
the MCP `expose` verb, the `EXPOSE_SET` job) and, since #648, they survive a
restart — but there was NO way to take one back. `Server.RemoveExposure` had no
caller outside tests, and `config.DeleteExposure` was written in #648 purely in
anticipation of this.

Concretely: an operator who shared Frigate by link had no supported way to stop
sharing it. Worse after #648 — the exposure is now durable, so the only
workaround (restart the worker) stopped working as one.

## Changes

- `internal/gateway/exposure.go`: `Server.Unexpose(name) bool` — drops the
  visibility policy AND clears the route's upstream address. Returns false if
  the name was not exposed, so a caller can say "not exposed" rather than
  claim a phantom teardown.
- `internal/status/agent.go`: `AgentProviders.Unexpose` + `POST /agent/unexpose`
  (same VPN-origin-or-org-token posture as `/agent/expose`; POST to match every
  other `/agent/*` route, so the backend needs no special-casing).
- `cmd/expose_ops.go`: `liveExposeOps.Unexpose` — tears down the live route
  first, then deletes the durable record.
- `cmd/service_unexpose.go`: `citadel service unexpose <name>`.
- `cmd/agent_tools.go`: wires the provider to the same live adapter as expose,
  so the CLI and any future job path cannot drift.
- `docs/gateway-ingress.md`: documents revoke and the ServeMux constraint.

## Two decisions worth reviewing

**Why clear the route as well as the policy.** Dropping the policy alone is
already fail-closed — `exposureMiddleware` 404s a name with no policy and is the
sole gate for that namespace. But the registered handler and its `Upstream`
would survive still pointing at the revoked service, so a later re-expose of the
same name would inherit the old target until it re-pointed it. Clearing the
address makes the stale route inert on its own terms (502, not a live backend).

**Why the Upstream is kept, not deleted.** `http.ServeMux` has no deregister, so
the mux entry outlives the revoke regardless. Deleting the `Upstream` from the
map would leave the registered handler closed over an orphan AND break
re-exposing: `wireExposeRoute` re-points an EXISTING entry in place, which is
exactly what makes a later expose of the same name work.

**Order: live route first, durable record second.** If the durable delete fails,
the service is already unreachable and the residue is a stale record that
resurrects on restart — noisy, but not an exposure the operator believes is
gone. The reverse order fails unsafely (record gone, service still serving,
nothing left to reconcile it). The error message says exactly that.

## Testing

`go test ./...` and `go vet ./...` pass.

New gateway tests assert the property that makes revoke real rather than
cosmetic — **the backend is never reached again**, not merely that the status
code is 404 (a policy-only revoke would pass a status-code-only test):
- backend hit count is unchanged after revoke, and the upstream address is empty
- an outstanding `link` token stops working (the shareable-credential case: the
  holder needs no mesh identity, so a missed revoke would keep working for a
  stranger)
- revoke is idempotent, and returns false for an unknown name
- re-expose after revoke works and lands on the NEW backend, not the old one

MUTATION TESTED both halves independently: disabling the route teardown fails
`TestUnexposeStopsReachingTheBackend`; disabling the policy delete fails three
tests. Neither half is redundant.

## Not included

No `EXPOSE_UNSET` job type and no aceteam-side MCP verb — the node side is what
was missing. `/agent/unexpose` is the endpoint the backend would call when an
MCP `unexpose` verb is added (aceteam repo, separate PR).

### DIFF
--- a/cmd/agent_tools.go
+++ b/cmd/agent_tools.go
@@ -87,6 +87,9 @@ func buildAgentProviders(ctx context.Context, d agentProviderDeps) *status.Agent
 				Epoch:      spec.Epoch,
 			})
 		},
+		Unexpose: func(name string) (any, error) {
+			return liveExposeOps{}.Unexpose(ctx, name)
+		},
 	}
 }
 
--- a/cmd/expose_ops.go
+++ b/cmd/expose_ops.go
@@ -84,6 +84,49 @@ func (liveExposeOps) Expose(_ context.Context, req worker.ExposeRequest) (*worke
 	return res, nil
 }
 
+// UnexposeResult is what the unexpose path returns to its caller (the CLI, the
+// aceteam MCP verb).
+type UnexposeResult struct {
+	// Name is the exposed-service slug that was revoked.
+	Name string `json:"name"`
+	// WasExposed reports whether a live exposure actually existed. False means
+	// the call still succeeded (revoke is idempotent) but nothing was serving --
+	// surfaced so a caller can say "not exposed" instead of implying it tore
+	// something down.
+	WasExposed bool `json:"was_exposed"`
+}
+
+// Unexpose revokes an exposure: it drops the gateway's live route AND the
+// durable record, so the service stops being reachable now and does not come
+// back on the next restart (#647 made exposures survive restarts, which is
+// exactly why revoke must clear both halves).
+//
+// Order matters. The LIVE route is torn down first: if the durable delete fails,
+// the service is already unreachable and the residual failure is a stale record
+// that resurrects on the next restart -- noisy but not an exposure the operator
+// believes is gone. Deleting the record first would invert that into the unsafe
+// direction (record gone, service still serving, nothing left to reconcile it).
+func (liveExposeOps) Unexpose(_ context.Context, name string) (*UnexposeResult, error) {
+	if name == "" {
+		return nil, fmt.Errorf("unexpose requires a service name")
+	}
+	ref := getProvisionedServiceGateway()
+	if ref == nil {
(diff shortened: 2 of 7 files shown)


sunapi386:
On-node verified on 1314 (mini), full expose → revoke → restart → re-expose lifecycle.

```
=== revoke ===
citadel service unexpose not-a-thing   -> "was not exposed; nothing to revoke."   (idempotent, no phantom teardown)
citadel service unexpose frigate       -> "Revoked"
exposures.json                          -> []                                      (durable record cleared)

GET /expose/frigate/  from an org peer -> 404   (was 200)
GET /                 from an org peer -> 200   (gateway itself unaffected)
curl localhost:8212/api/version         -> 200   (revoke unshares; it does NOT stop the service)

=== stays revoked across a restart (the inverse of #648) ===
systemctl restart citadel-worker       -> no "Restored ... exposure" line; nothing came back

=== re-expose works, on the kept-Upstream path ===
citadel service expose frigate --port 8212 --visibility org
GET /expose/frigate/  from an org peer -> 200, assets rewritten to /expose/frigate/assets/main-*.js
```

One measurement worth flagging: my first check grepped the last `Restored` line and saw one, which reads like the exposure came back. It had not — timestamps showed the line was `07:57:13`, from the restart BEFORE the revoke at `07:57:23`; the post-revoke restart at `07:57:50` logged nothing. Confirmed against `journalctl` start times rather than the grep.

Node has been returned to the released **v2.97.0** binary (this PR is unreleased), exposure restored, `/expose/frigate` 200.

PR #665, opened 2026-08-02
feat(meeting): bot->room virtual-mic speaking path

sunapi386:
## Context

Part of aceteam#7079 (real-time two-way conversation in meetings). Today the
containerized meeting bot can **HEAR** a meeting (room->bot capture via a
PulseAudio null-sink monitor recorded by ffmpeg) but cannot **SPEAK** into it —
there is no path to inject audio as the Chromium tab's microphone. This adds the
mirror path: a virtual microphone plus a `meetingd` control endpoint to inject
audio into it. Strictly additive and gated, so a bot that never speaks behaves
exactly as today.

## Root Cause

The meeting-service container (`services/meeting-service/`) had only a capture
null sink. There was no virtual *source* the tab could publish as its mic, and no
control path to feed audio into one.

## Changes

- **`citadel.pa`** — boot a `citadel_mic` null sink whose monitor is **remapped**
  (`module-remap-source`, `device.class=sound`) to a real source `citadel_virtmic`,
  set as the container's default input device so the Chromium tab publishes it as
  its mic. Guarded by `.nofail` (load-bearing: `pulseaudio -n -F` is fail-fast by
  default, so a mic-line failure would otherwise EXIT the daemon and kill capture
  too — `.nofail` degrades it to "mic absent"). Additive: the capture sinks and the
  `/health` canary name `<sink>.monitor` explicitly, so the new default source
  cannot touch them.
- **`meetingd.py`** — two speaking endpoints (one body shape each):
  - `POST /mic/play` — JSON `{"path": "<workspace-relative audio file>"}`, any
    ffmpeg-decodable format, path-safety via the same `_safe_workspace_path` guard
    the recorder uses; decoded and played into `citadel_mic` via `paplay`.
  - `POST /mic/play/pcm` — raw signed-16-bit LE PCM body, `?rate=&channels=`
    (default 24000/1), streamed via `pacat`.

  Both are node-wide, serialized (`_mic_lock`, 409 `already speaking`), and
  **synchronous**. `/health` now reports `virtual_mic:{present,sink,source,required}`
  and only 503s on an absent mic when `MEETING_MIC_REQUIRED` is set (default:
  reported, never fails capture).
- **`compose.yml`** — pass through `MEETING_MIC_REQUIRED` (compose does not forward
  host env vars that aren't listed, so the documented opt-in needs this).
- **`internal/jobs/meeting_media.go`** — `containerMedia.SpeakFile` helper POSTs to
  `/mic/play` with its own 3-min client (a clip outlasts the 30s control client).
  Additive/unused until the realtime engine is wired in a later wave.
- **Tests** — 12 new `meetingd` tests (pure arg builders, health report-not-fail
  contract, mic-play guards + happy paths, `citadel.pa`/`.nofail` pins) and 2 Go
  `SpeakFile` httptest cases. `smoke_test.py` step 6 live-validates the path. README
  + CLAUDE.md documented.

## Testing

Ran from the worktree root:

- `go build ./...` — pass (exit 0)
- `go vet ./...` — pass (exit 0)
- `go test ./...` — pass (all packages `ok`; new `TestContainerMediaSpeakFile` +
  `...ErrorStatus` pass)
- `python3 -m py_compile services/meeting-service/{meetingd,test_meetingd,smoke_test}.py` — OK
- `pytest services/meeting-service/test_meetingd.py` — **26 passed**
- `ruff check` — flags only the same PLW1510/SIM115/F541 categories already present
  in the pre-existing files (no repo ruff config is enforced on this service).

**Not runnable here (no Docker on this box):** the meeting-service image was not
built. Live validation on a node (meeting container runs on node 1297; CDP host
8208, meetingd host 8207):

1. `python3 services/meeting-service/smoke_test.py` (needs Docker + `websocket-client`).
   Step 6 grants mic permission, confirms `citadel_virtmic` enumerates as an
   `audioinput`, then asserts a `getUserMedia` analyser HEARS a tone POSTed to
   `/mic/play`. **A step-6 failure at the `getUserMedia`/`started` assert means
   "mic permission not granted, not mic broken"** (the browser-domain
   `Browser.grantPermissions` over a page target can be inconsistent) — the human
   fallback is a profile-level mic grant for the origin.
2. **Live-validation #1:** exec into the container and confirm
   `pactl list sources` shows `citadel_virtmic` with **no** `Monitor of Sink:`
   line and `device.class = sound` — this single fact decides whether Chromium
   selects it as a mic. If it inherits the monitor class, Chromium filters it out.
3. `curl -sS -XPOST 127.0.0.1:8207/health | jq .virtual_mic` → `present:true`.

## What is NOT done / known risks

- **Mic permission grant** is not wired: `build_chrome_args` deliberately keeps the
  pinned flag set (it's the capture path's tested contract), so no
  `--use-fake-ui-for-media-stream`. The tab is granted mic access only via the
  persisted profile's prior grant or a CDP `Browser.grantPermissions`. Next
  required piece.
- **No realtime-engine WS wiring** and **no barge-in / mid-clip stop** (playback is
  synchronous, one clip at a time) — the later wave.
- The monitor-vs-source classing (item 2 above) is the #1 live-validation item;
  `device.class=sound` is the intended fix but is unverified without a container.

## Endpoint request contracts

- `POST /mic/play` — `Content-Type: application/json`, body `{"path":"<ws-rel>"}`.
  → 200 `{"played":true,"source":"file","path":"<abs>"}`; 400 bad/traversal path;
  404 missing file; 409 already speaking; 503 pulse down / mic absent; 504 timeout.
- `POST /mic/play/pcm` — raw `application/octet-stream` body = s16le PCM; query
  `rate` (default 24000), `channels` (default 1). → 200
  `{"played":true,"source":"pcm","bytes":N}`; 400 empty body / non-positive params;
  409 already speaking; 503 pulse down / mic absent; 504 timeout.

Closes part of aceteam#7079.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -711,6 +711,35 @@ shaped like:
 The `engine` must be `bonsai` so the node routes `MODEL_CACHE_PULL` to the
 single-file GGUF pull and `SERVICE_START` to `services/compose/bonsai.yml`.
 
+### Meeting media stack — virtual mic / bot speaking path (aceteam#7079)
+
+The `meeting-service` module (`services/meeting-service/`) is a headed Chromium +
+Xvfb + PulseAudio container (`meetingd` control API on host 8207 / container 8102,
+CDP on host 8208) that lets the bot HEAR a meeting via a null sink whose monitor
+ffmpeg records. The bot->room SPEAKING path adds a **virtual microphone**:
+`citadel.pa` boots a `citadel_mic` null sink whose monitor is **remapped**
+(`module-remap-source`, `device.class=sound`) to a real source `citadel_virtmic`,
+set as the default input device so the Chromium tab publishes it as its mic.
+
+- **Inject audio** via meetingd: `POST /mic/play` (JSON `{"path": "<ws-rel file>"}`,
+  any ffmpeg format → `paplay`) or `POST /mic/play/pcm` (raw s16le body,
+  `?rate=&channels=` → `pacat`). Node-wide, serialized (409 `already speaking`),
+  synchronous (returns when the clip finishes); no barge-in yet. Go transport:
+  `containerMedia.SpeakFile` (own 3-min client, not the 30s control client).
+- **Strictly additive & gated:** capture sinks and the `/health` canary name
+  `<sink>.monitor` explicitly, so making `citadel_virtmic` the default *source*
+  can't touch capture. `/health` **reports** `virtual_mic:{present,sink,source,
+  required}` but only 503s on absence when `MEETING_MIC_REQUIRED=1` — a node that
+  never speaks behaves exactly as before.
+- **Load-bearing subtlety:** Chromium filters *monitor*-class sources out of
+  `getUserMedia`, which is why the monitor is remapped to a `device.class=sound`
+  source rather than used raw. Live-validation #1: `pactl list sources` must show
+  `citadel_virtmic` with **no** `Monitor of Sink:` line (see the meeting-service
+  README + `smoke_test.py` step 6).
+- **NOT done (next wave):** mic-permission grant (`build_chrome_args` has no
+  `--use-fake-ui-for-media-stream`; rely on the profile's prior grant or a CDP
+  `Browser.grantPermissions`), realtime-engine WS streaming, and barge-in.
+
 ### Per-request Energy Receipt (footprint energy, aceteam#6635)
 
 The footprint sampler can record a per-interval node energy estimate so the
--- a/internal/jobs/meeting_media.go
+++ b/internal/jobs/meeting_media.go
@@ -152,6 +152,12 @@ const (
 	meetingContainerCDPTimeout = 30 * time.Second
 	// meetingContainerHTTPTimeout bounds a single meetingd control call.
 	meetingContainerHTTPTimeout = 30 * time.Second
+	// meetingSpeakTimeout bounds a blocking POST /mic/play. meetingd plays the clip
+	// SYNCHRONOUSLY (it returns when playback finishes), and a TTS clip can run tens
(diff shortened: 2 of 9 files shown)


sunapi386:
**Live-validated the #1 risk (monitor-vs-source classing) on node 1297's running meeting container** (PulseAudio protocol 35, image `:latest`), by loading the exact `citadel.pa` module pair against the live daemon (isolated test names, unloaded after — container left as found):

\`\`\`
load-module module-null-sink sink_name=citviptest ...
load-module module-remap-source master=citviptest.monitor source_name=citvirtmictest source_properties=device.class=sound
\`\`\`
Result for the remapped source:
\`\`\`
Name: citvirtmictest
Monitor of Sink: n/a          # ← real source, NOT a monitor (a true monitor names a sink here)
Flags: DECIBEL_VOLUME LATENCY
device.class = "sound"        # ← classed as a real input device
\`\`\`
→ **PASS**: `module-remap-source` + `device.class=sound` yields a non-monitor, sound-classed capture source, which is what makes it visible to Chromium's `getUserMedia` enumeration. (`Description: "Remapped Monitor of ..."` is a cosmetic label only; the load-bearing fields are `Monitor of Sink: n/a` + `device.class=sound`.)

**Still pending for un-draft:** full end-to-end on a build of THIS image — Chromium actually selects `citadel_virtmic` as default mic (`set-default-source`) and `/mic/play` publishes into a live meeting. The released `:latest` predates this PR, so that step needs an image build (deferred; heavy Chromium/Xvfb/PulseAudio image). Static + unit gate is green (`go build/vet/test`, 26 pytest, py_compile).

Issue #690, opened 2026-08-05
Embedding services report no models, so a stopped engine claims their model


sunapi386:
## What is broken

`collectEmbeddingServiceStatus` (`internal/status/engines.go:120-148`) never populates `Models`. So a running `tei` embedding engine reports no models array at all, while a stopped `vllm` entry claims `Alibaba-NLP/gte-multilingual-base` from its `<name>.env` default.

Net effect upstream: an embedding model is advertised as a chat model, attributed to a dead engine, while the live engine that actually serves it advertises nothing.

## Why it matters on constrained hardware

Embedding servers are small and cheap to keep resident, which makes them exactly the thing a constrained box should pin rather than swap. Mis-attributing their models to a large stopped engine means the scheduler reasons about the wrong VRAM cost and the wrong lifecycle for both.

## Suggested fix

Populate `Models` in `collectEmbeddingServiceStatus` from the engine's own listing, and stop letting a stopped engine's env-file default claim a model another engine is serving.

Part of the serving stack epic aceteam-ai/aceteam#7243.

Issue #632, opened 2026-07-30
feat: native multi-model residency — VRAM-aware on-demand model hotswap


sunapi386:
## Problem
A single GPU node fits only one large model at a time (e.g. node 1297 / RTX 3090: Unlimited-OCR ~13GB **or** Bonsai-27B ~21GB, not both). Today deploying a second large model forces evicting the first, so the fabric can only offer one big model per node at a time. We want a node to **natively present multiple installed models as available and swap them in on demand**, instead of an operator picking a single resident model.

## Design: VRAM-aware on-demand hotswap
Build on the primitives that already exist:
- **#577** `PlanPreemption(candidates, requiredVRAM, availableVRAM)` (`internal/status/preempt.go`) + `pinned_services` — the eviction planner.
- **#416** `AutoStopReconciler` (`internal/status/autostop.go`) — idle auto-stop.
- Catalog VRAM (`fabric_provision.resolve_model_vram` / catalog `vram_gb`) — the per-model budget.
- Installed-vs-resident precedent: ollama already reports installed (not just resident) models (`internal/status/models.go:68`).

**Behavior:** the node advertises all *installed* models as available + which are *resident*. On an inference request for an installed-but-not-resident model, the node preempts idle/LRU non-pinned resident models (PlanPreemption, budget = catalog VRAM), activates the requested model, and serves. Idle models auto-stop to keep headroom; pinned models never swap out.

## Hard constraints (these decide whether it works)
1. **Cold-start latency IS the feature.** A swap-in on the request path blocks for the model load. Observed this session: Unlimited-OCR first-ever load (weights download) = **several minutes**; cached warm reload = tens of seconds to ~1-2 min. The gateway has a **120s `WriteTimeout`** and `run_fabric_chat` is a buffered dispatch — so a swap MUST NOT hold the socket through a vLLM load. **Contract:** on a residency miss, return a fast structured `model_warming` response (with an ETA) and let the caller retry; only serve once resident. (Measure real bonsai↔OCR swap time on node 1297 and pin the number in phase 1.)
2. **Single-flight swap lock + min-residency floor.** Once the node advertises installed-not-resident models, `resolve_fabric_target` will route mid-swap; concurrent requests for different models must not thrash the GPU. Serialize swaps per node and forbid evicting a model within N seconds of loading.
[...]

PR #702, opened 2026-08-05
fix(heartbeat): report the services that are actually running, not the ones we thought to probe

sunapi386:
## Context

aceteam-ai/aceteam#7148. Deploying `gte-multilingual-base` to node `1314` started TEI successfully, but the node's own reported inventory did not show it:

| Call | Result |
| --- | --- |
| `service_status(1314, "tei")` | `tei is running (docker)` |
| `fabric_node_status(1314)` | `Running Services: none` |

The direct `SERVICE_STATUS` dispatch is authoritative, so the node was telling the platform two different things about the same container.

## Root cause

Not "the inventory enumerates manifest-declared services" as the issue guessed. The manifest list (`c.services`) is nil on the heartbeat path, and the `stopped` rows an operator sees come from `applyModelHotswap` (#632) advertising installed-but-stopped engines. The real defects are both in `internal/status/engines.go`:

1. **Coverage.** Only two hardcoded lists were ever consulted: `managedProbeEngines` (vllm, ollama, llamacpp, bonsai, unlimited-ocr) and `embeddingProbeServices` (tei). `kokoro`, `transcribe`, `diffusers`, `extraction`, `sglang` and `lmstudio` had no path into the heartbeat at all, running or not.

2. **Readiness gate.** Even for a listed service, an entry was emitted only if a probe answered (`if !responded { continue }`) or TEI's `/health` returned 200. A container that is up but still downloading weights was omitted **entirely** rather than reported as not-ready. That is the exact repro: TEI was pulling ~600MB of `gte-multilingual-base`, so it reported nothing for the whole warm-up. Node 1297 listed `tei` only because its TEI had finished warming long ago.

Because #6909's TEI provisioning uses `SERVICE_START` from the embedded `ServiceMap`, this hit every embedded-compose engine started through Quick Deploy or the per-node Deploy tab.

## Changes

- `internal/status/running_services.go` (new): enumerates running `citadel-*` containers with ONE `docker ps` (or podman, via `catalog.SelectContainerRuntime()`) per heartbeat and maps them back to `services.ServiceMap` keys. The container lister is a package var so the selection logic is unit-testable without a container runtime. Replaces the previous one `docker inspect` per known engine.
- `internal/status/engines.go`: both probe passes now report a running-but-unresponsive service with `health=starting` instead of dropping it. `status` stays `running` (the container really is running, which is what `SERVICE_STATUS` reports) and `health` carries readiness. Readiness gating moves from presence to health.
- Backstop pass reports every other running embedded-compose service. `Type` is deliberately conservative: only genuine OpenAI chat engines get `llm`, so kokoro/transcribe/diffusers/extraction show up in the inventory without becoming chat-router candidates.
- `internal/status/types.go`: `HealthStatusStarting`.

## Rebased onto main (v2.99.0), scope reduced

This branch was rebased onto `origin/main` after #699 landed. Two files conflicted, `internal/status/engines.go` and `internal/status/models.go`, and the resolution dropped one of this PR's original contributions:

- **TEI model discovery is no longer part of this PR.** `beaa29a9` (citadel-cli#690) already added `ModelDiscovery.DiscoverEmbeddingModel`, reading the served model id from TEI's `GET /info`, and deliberately kept it OFF the `DiscoverModels` chat-engine switch. This branch's `models.go` change added a `case "tei"` to that very switch, which would have reintroduced the mis-attribution #690 fixed. `models.go` is reverted to main's version and this PR now calls `DiscoverEmbeddingModel`.
- `collectEmbeddingServices` keeps #690's injected running/health seams. Its signature narrows from `(ctx, engineBin, func(engineBin, name string), healthy, md)` to `(ctx, func(name string), healthy, md)`: the production wrapper now supplies a closure over the once-per-heartbeat running set, so `engineBin` is no longer consulted inside. The two upstream call sites in `embedding_models_test.go` are updated. The native-install fallback #690 added is preserved, via `enginePortIfRunning`, so a natively served embedding server is still reported.
- This PR's own `embeddingHealthProbe` package var is removed in favour of upstream's injected `healthy` parameter, so there is one test seam for readiness rather than two.

## Deploy ordering (satisfied)

The platform half, aceteam-ai/aceteam#7263, shipped in **v1.359.0** and is deployed. It adds the readiness gate this change depends on: `utils/rag_embeddings._service_is_tei` matches on service NAME only, so a `starting` TEI reported without that gate would have made a warming node eligible for embedding dispatch and the embed would 503.

## Behaviour note

During an engine warmup, a managed engine now reports `running` + `starting` with no models instead of being absent and picked up by `collectInstalledEngines` as `stopped` with its `<name>.env` model. So for the warmup window that model id is not in the node's advertised set, and `applyModelHotswap` marks the warming engine `Resident=true` with a VRAM estimate. Both are more honest than the old shape: the card really is being occupied, and the model really is not servable yet. No duplicate entry is possible, `collectInstalledEngines` skips by name against `reported`.

## Testing

```
gofmt -l .   # clean
go vet ./... # clean
go test ./... # 68 packages pass
```

New tests:

- `running_services_test.go`: container-name mapping, module containers excluded, runtime failure degrades to empty, EVERY `ServiceMap` entry is reportable, already-reported services are not overwritten, deterministic ordering, and `llm` is reserved for chat engines.
- `starting_service_test.go`: a warming TEI is reported `running` + `starting` with no models; a ready TEI is `ok` and reports its served model id; an unresponsive vLLM is reported `starting` with no idle signal; the collector wrapper threads the running set into that same decision; and `managedProbeEngines` + `embeddingProbeServices` are subsets of `services.ServiceMap` (the new enumeration filters on `ServiceMap` membership, which the per-engine `docker inspect` it replaced did not).

Coverage is defined by `services.ServiceMap`: ollama, vllm, llamacpp, lmstudio, sglang, extraction, transcribe, diffusers, bonsai, kokoro, tei, unlimited-ocr. `transcribe` is the whisper service; there is no `comfyui` entry, `diffusers` is the image service.

Not verified on a live node: nodes 1297 and 1314 carry demo work for an Aug 14 customer meeting and were left untouched.

### DIFF
--- a/internal/status/collector.go
+++ b/internal/status/collector.go
@@ -133,7 +133,15 @@ func (c *Collector) Collect() (*NodeStatus, error) {
 	for _, svc := range status.Services {
 		reported[svc.Name] = struct{}{}
 	}
-	for _, eng := range c.collectManagedEngineStatus() {
+
+	// Enumerate the running "citadel-<name>" containers ONCE and thread the set
+	// through every service collector below (aceteam#7148). One container-runtime
+	// call per heartbeat instead of one per known engine, and, the reason it
+	// exists, a single source of truth for "what is actually up" that no longer
+	// depends on a service appearing in a hardcoded probe list.
+	runningServices := runningEmbeddedServices(containerRuntimeBin())
+
+	for _, eng := range c.collectManagedEngineStatus(runningServices) {
 		if _, dup := reported[eng.Name]; dup {
 			continue
 		}
@@ -145,12 +153,22 @@ func (c *Collector) Collect() (*NodeStatus, error) {
 	// is the discovery signal the sovereign-RAG backend matches on to embed on
 	// the org's own node; like the engine probe it runs in the heartbeat path
 	// where c.services is nil, so a manifest entry is not required.
-	for _, emb := range c.collectEmbeddingServiceStatus() {
+	for _, emb := range c.collectEmbeddingServiceStatus(runningServices) {
 		if _, dup := reported[emb.Name]; dup {
 			continue
 		}
 		reported[emb.Name] = struct{}{}
 		status.Services = append(status.Services, emb)
+	}
+
+	// Backstop: every OTHER running embedded-compose service (kokoro, transcribe,
+	// diffusers, extraction, sglang, lmstudio). None of these has a probe above,
+	// so before #7148 a node running them reported nothing at all: the operator
+	// saw an empty node right after a successful deploy, and an unreported
+	// service can never be flagged idle or reclaimable either.
+	for _, svc := range collectRunningEmbeddedServices(runningServices, reported) {
+		reported[svc.Name] = struct{}{}
+		status.Services = append(status.Services, svc)
 	}
 
 	// Mark pinned services (citadel #577) so the heartbeat and `citadel services`
--- a/internal/status/embedding_models_test.go
+++ b/internal/status/embedding_models_test.go
@@ -90,7 +90,7 @@ func TestCollectEmbeddingServices_NativeProcessCounts(t *testing.T) {
 	// portIfRunning stands in for managedEnginePortIfRunning, whose real
 	// implementation checks the container first and falls back to the native
 	// service probe. Here only the native branch answers.
-	nativeOnly := func(_, name string) (int, bool) {
+	nativeOnly := func(name string) (int, bool) {
 		if name == "tei" {
 			return 8102, true
 		}
@@ -98,7 +98,7 @@ func TestCollectEmbeddingServices_NativeProcessCounts(t *testing.T) {
 	}
 	md := stubEmbeddingLister{models: []string{"Alibaba-NLP/gte-multilingual-base"}}
 
-	got := collectEmbeddingServices(context.Background(), "docker", nativeOnly, alwaysHealthy, md)
+	got := collectEmbeddingServices(context.Background(), nativeOnly, alwaysHealthy, md)
 	if len(got) != 1 {
 		t.Fatalf("expected the native embedding server reported, got %+v", got)
 	}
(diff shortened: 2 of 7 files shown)


sunapi386:
CI note: run [31028872723](https://github.com/aceteam-ai/citadel-cli/actions/runs/31028872723) is green on `383b751` (the functional commit). GitHub created no workflow run for the follow-up commit `d26b69c`, despite the synchronize push and two close/reopen cycles, so the PR currently shows no checks.

`d26b69c` is comment-only:

```
$ git diff 383b751 d26b69c --stat
 internal/status/collector.go             | 6 +++---
 internal/status/engines.go               | 2 +-
 internal/status/running_services.go      | 8 ++++----
 internal/status/starting_service_test.go | 2 +-
 internal/status/types.go                 | 2 +-
 5 files changed, 10 insertions(+), 10 deletions(-)

# every changed line is a // comment; non-comment diff is empty
```

Verified locally on `d26b69c`: `gofmt -l internal/status/` clean, `go vet ./...` clean, `go test ./...` all packages pass. Worth re-running CI before merge.

sunapi386:
**Reviewer note (documentation, not a code change):** this PR widens the set of engines that appear in `status.Services` with an attached `Footprint`/`IdleState` to include the backstop-detected engines (diffusers, kokoro, transcribe, extraction, sglang, lmstudio). A side effect is that these engines now become eligible **VRAM-preemption candidates** (`buildPreemptCandidates`/`preemptForVRAM`) and **`SERVICE_AUTO_STOP_WHEN_IDLE` eviction candidates** — neither was possible before, since those entries didn't exist in `status.Services`. Both mechanisms are currently gated (VRAM-budget forwarding is inert; auto-stop is default-OFF), so this is latent, not an active behavior change. But a node that already has `SERVICE_AUTO_STOP_WHEN_IDLE=true` set will begin evicting engines it previously couldn't touch. Flagging so it's a reviewed decision rather than a silent side effect.

Issue #741, opened 2026-08-11
reconnect can dial after Close because done is not re-checked after the backoff sleep


sunapi386:
## What is broken

`WSClient.reconnect` checks `c.done` at the top of its loop and then sleeps for
the current backoff, which can be up to `maxBackoff` (one minute). It does not
re-check `c.done` after the sleep, so it goes straight into `connectLocked` and
dials.

`internal/redisapi/websocket.go`, in `reconnect`:

```go
select {
case <-c.done:
    return
default:
}
... // connected check
time.Sleep(delay)   // up to maxBackoff
... // no done re-check here
c.connMu.Lock()
err := c.connectLocked(context.Background())
c.connMu.Unlock()
```

## Consequence

`Close` closes `done` and closes the connection, but a reconnect goroutine that
was already sleeping will still dial afterwards. On success it sets
`c.connected = true` and `c.conn = <new conn>` on a client the caller has
already closed, so `IsConnected()` reports true after `Close` returned, and a
socket is left open until the next loop iteration notices `done` and returns
without closing it.

The window is the full remaining sleep, so it is largest exactly when the
control plane is unhealthy and the backoff has grown.

Two smaller things in the same lines:

- `connectLocked` is called with `context.Background()`, so the dial is not
  bound by the caller's lifetime at all. The 10s handshake timeout is the only
  bound.
- `time.Sleep` is not cancellable, so `Close` waits out the backoff instead of
  returning promptly. `cmd/ws_pubsub.go` deliberately uses ctx-aware chunked
  sleeps for the same reason; this loop predates that.

## Suggested fix

Replace the bare `time.Sleep` with a select on a timer and `c.done`, and
re-check `c.done` after it. Consider giving the client a context so
`connectLocked` gets something better than `context.Background()`.

## Related

- #728 unsynchronized `reconnectBackoff`, same function
- #723 / PR #732, which established the ctx-aware chunked sleep pattern

sunapi386:
Fixed on the integration branch in PR #749.

`time.Sleep(delay)` is replaced by `sleepUnlessDone`, a select on a timer and
`c.done`, so `Close` during a backoff both returns promptly and prevents the
dial that used to follow. It is fixed there rather than deferred because the
#734 keepalive changes its odds: `reconnect` used to run almost never and now
runs on every read-deadline expiry, so this window opens routinely instead of
rarely, and the reconnect jitter added in the same PR lengthens the sleep the
[...]

Issue #740, opened 2026-08-11
Connect during a reconnect starts a second readLoop on the same gorilla Conn


sunapi386:
## What is broken

`WSClient.Connect` unconditionally starts a read loop:

```go
if err := c.connectLocked(ctx); err != nil {
    return err
}
go c.readLoop()
```

`reconnect` calls the same `connectLocked` but does NOT start one, because it
assumes the read loop started by the original `Connect` is still alive and will
pick up the new `c.conn` on its next iteration.

That assumption breaks when `Connect` runs while a reconnect is in flight. Then
there are two read loops, both re-reading `c.conn` every iteration, and both
calling `conn.ReadMessage()` on whichever connection is current.

gorilla/websocket permits exactly ONE concurrent reader, the same way it permits
exactly one concurrent writer. Two concurrent readers corrupt the shared frame
state on `Conn` (`readErr`, `readRemaining`, `readFinal`, the reader chain). This
is the same class of contract violation as #720, which was the write side.

## Why it is reachable

`Client.EnableWebSocket` is documented as idempotent and safe to call
repeatedly. It returns early only when `IsConnected()` is true, and
`IsConnected()` is false for the whole reconnect window, which is up to
`maxBackoff` (one minute). So any `EnableWebSocket` call that lands in that
window forwards to `Connect`, which dials a second connection and starts a
second read loop.

`EnableWebSocket` has several callers that can overlap:

- `cmd/ws_pubsub.go` `enableWebSocketWithRetry`, whose background loop retries on
  a schedule until it succeeds (#723)
- `internal/worker/ws_source.go` `WSSource.Connect`
- `internal/chat/client.go` `Client.Connect`
- `cmd/controlcenter.go`

## Evidence

Observed under `-race` while writing the regression test for #728. With the
racing goroutine calling the exported `Connect` rather than `connectLocked`,
the detector reports:

```
WARNING: DATA RACE
Write at ... by goroutine 130:
  github.com/gorilla/websocket.(*Conn).NextReader()
  github.com/gorilla/websocket.(*Conn).ReadMessage()
  ...redisapi.(*WSClient).readLoop()
  ...redisapi.(*WSClient).Connect.gowrap2()

Previous write at ... by goroutine 140:
  (identical stack, different goroutine)
```

Both read loops were started by `Connect` at `internal/redisapi/websocket.go:148`.

## Suggested fix

Make connection ownership explicit rather than implicit. Options, roughly in
increasing order of invasiveness:

[...]

Issue #723, opened 2026-08-10
A single failed WebSocket connect at startup disables pub/sub permanently and silently


sunapi386:
## What is broken

`EnableWebSocket` is called exactly once, at worker startup. If that single
connect attempt fails, the client is discarded and never retried for the entire
lifetime of the process, and the failure is logged only at Debug level.

`cmd/work.go`:

```go
// Enable WebSocket for real-time pub/sub (heartbeat, streaming responses)
if err := apiSource.Client().EnableWebSocket(ctx); err != nil {
    // WebSocket is optional - fall back to HTTP
    Debug("WebSocket not available, using HTTP for pub/sub: %v", err)
}
```

`internal/redisapi/client.go`:

```go
c.wsClient = NewWSClient(WSClientConfig{..., ReconnectEnabled: true, ...})
if err := c.wsClient.Connect(ctx); err != nil {
    c.wsClient = nil
    return fmt.Errorf("failed to connect WebSocket: %w", err)
}
```

`ReconnectEnabled: true` only governs reconnection of a connection that was
established at least once. When the *initial* `Connect` fails, `wsClient` is set
back to `nil`, so there is no object left to reconnect and nothing schedules
another `EnableWebSocket`.

## Why this matters more than "WebSocket is optional"

The fallback it degrades to does not work. `Publish` chooses HTTP when
`wsClient` is nil, and the HTTP publish path always reports failure because it
reads a response field the route does not send (separate issue). So "optional
with an HTTP fallback" is, in practice, "pub/sub is permanently dead for this
process."

Because the log line is `Debug`, a node in this state produces no evidence at
default verbosity that the WebSocket was ever attempted. Diagnosing it required
reading the call site; the journal shows nothing at all.

## Observed

A node restarted at 23:39:18 while the control plane was mid-deploy (the
preceding minutes show `502 upstream error` and client timeouts against
`aceteam.ai`). The startup `EnableWebSocket` failed into the Debug branch.
Twelve hours later, with the control plane long healthy, the process was still
using the HTTP path and had never retried, so the node had published no durable
heartbeat that whole time and the platform considered it offline.

The timing is the ordinary case, not bad luck: workers restart *because* the
control plane blipped, so the startup connect is disproportionately likely to
land inside the very window that breaks it.

## Suggested fix

Retry the WebSocket connect in the background with the same
backoff-and-jitter treatment `connectWithBackoff` already applies to the
control-plane connect a few lines above, rather than giving up after one try.
[...]

Issue #720, opened 2026-08-10
WSClient.sendMessage writes under a read lock, so concurrent writes panic and kill the worker


sunapi386:
## What happened

A node running 2.100.0 panicked and died:

```
panic: concurrent write to websocket connection
github.com/gorilla/websocket.(*messageWriter).flushFrame
github.com/gorilla/websocket.(*Conn).WriteMessage
github.com/aceteam-ai/citadel-cli/internal/redisapi.(*WSClient).sendMessage
        internal/redisapi/websocket.go:418
github.com/aceteam-ai/citadel-cli/internal/redisapi.(*WSClient).Subscribe
        internal/redisapi/websocket.go:349
github.com/aceteam-ai/citadel-cli/internal/redisapi.(*WSClient).connectLocked.func1()
        internal/redisapi/websocket.go:177
created by ...connectLocked in goroutine 847
```

systemd restarted the worker 39 seconds later. The restart is not the problem;
what the crash leads to is (see "Consequence").

## Root cause

`WSClient.sendMessage` (`internal/redisapi/websocket.go`) takes a **read** lock
to guard the connection pointer, then performs the write while still holding
only that read lock:

```go
c.connMu.RLock()
conn := c.conn
connected := c.connected
c.connMu.RUnlock()
...
if err := conn.WriteMessage(websocket.TextMessage, data); err != nil {
```

`RLock` is shared, so any number of goroutines can be inside `WriteMessage`
concurrently. gorilla/websocket documents that only one concurrent writer is
permitted and panics otherwise. There is no write mutex anywhere in the file.

The writers really are concurrent: the heartbeat publisher ticks every 30s
(`APIPublisher.publishStatus` -> `Publish` -> `sendMessage`), job streaming
calls `PublishStreamEvent` per chunk, and `connectLocked` spawns a goroutine
that calls `Subscribe` -> `sendMessage` on reconnect. The reconnect path racing
the heartbeat is exactly what the stack above shows.

## Consequence

The panic is raised on a goroutine, so it takes the whole worker process down
rather than failing one publish. On restart the WebSocket does not come back,
and the node silently falls back to the HTTP publish path, which is itself
broken (see the linked issues). Net effect on the node that hit this: no
durable heartbeat for 12+ hours while the worker was otherwise healthy and
consuming jobs normally.

## Suggested fix

Add a dedicated write mutex held across `WriteMessage` (a separate `writeMu
[...]

PR #743, opened 2026-08-11
fix(redisapi): guard reconnectBackoff with connMu (#728)

### LINKED ISSUE #728: reconnectBackoff is written under connMu.Lock in one place and unlocked in another
## What is broken

`reconnectBackoff` on `WSClient` is written under `connMu.Lock` in one place and
read/written with no lock at all in another.

- `connectLocked` (around line 161) writes it while holding `connMu.Lock`.
- `reconnect` (around lines 296-303) reads and updates it unlocked.

That is a data race by the Go memory model, independent of the WebSocket write
race fixed in #720/#725. Found while adding the write mutex for #720; deliberately
left out of that PR to keep the diff to one concern.

## Consequence

Torn or stale reads of the backoff duration, so the reconnect schedule can be
wrong: too aggressive is the direction that matters. citadel already has history
here, see the comment above `connectWithBackoff` in `cmd/work.go` about #443,
where a restart storm burned the node's daily Redis-API quota and hard-locked
the node for roughly 24 hours. Anything that can silently shorten a backoff
deserves to be correct rather than approximately correct.

## Suggested fix

Guard `reconnectBackoff` consistently, either under `connMu` at every access or
behind its own small mutex, and add a `-race` test that exercises a reconnect
concurrently with a connect.

Note that #723 is changing reconnect and connect behaviour (adding retry with
backoff for a failed first connect), so these two should be sequenced rather
than developed in parallel against the same lines.

## Related

- #720 / PR #725 WebSocket concurrent write panic
- #723 one-shot connect never retried

sunapi386:
## Context

Closes #728. `reconnectBackoff` was written under `connMu` in `connectLocked`
and read and advanced with no lock at all in `reconnect`.

The race is real on current `main`, and it reproduces under `-race`. It does not
reproduce the way the issue thread describes, so the mechanism is worth
correcting before the fix.

## The race as described, versus the race that is there

The escalation comment on #728 says PR #725 lets the read path and the write
path start two `reconnect()` goroutines concurrently, and that the second one
"bails at `alreadyConnected`". On current `main` the second caller never reaches
`go c.reconnect()` at all. `handleDisconnect` takes `connMu`, the first caller
sets `c.conn = nil`, and the second caller then fails the connection-pointer
guard (`conn != nil && c.conn != conn`) and returns one layer earlier.

So a single disconnect event cannot spawn two reconnects, which is what the
read-path/write-path pair in #725 would produce. Two reconnects can still
coexist, but only by a longer route: reconnect#1 is sleeping, a `Connect()`
lands in that window and succeeds, the connection it established then fails, and
`handleDisconnect` spawns reconnect#2 while #1 is still alive. That route needs a
`Connect()` inside the reconnect window, which is the #723 precondition below,
not anything #725 introduced.

The pair that actually collides is `Connect()` versus an in-flight `reconnect()`:

- `connectLocked` writes `reconnectBackoff` at websocket.go:190 under `connMu`
- `reconnect` reads and advances it at websocket.go:334-341 under nothing

What makes that pair ordinary rather than exotic is #723 / PR #732, not #725.
`Client.EnableWebSocket` is documented as idempotent and safe to call
repeatedly; it returns early only when `IsConnected()` is true, and
`IsConnected()` is false for the entire reconnect window, which is up to
`maxBackoff` (one minute). `enableWebSocketWithRetry` now keeps calling it in the
background on a schedule until it succeeds, and `worker/ws_source.go`,
`chat/client.go` and `controlcenter.go` call it too. Any of those landing inside
a reconnect window forwards straight to `Connect`.

The consequence is the one #728 flags: a stale read of the backoff shortens the
retry schedule, which is the direction that produced the #443 restart storm.

## Reproduction

`TestReconnectBackoffAccessIsSynchronized` fails under `-race` on `main`:

```
WARNING: DATA RACE
Write at 0x00c00018a2b0 by goroutine 16:
  ...redisapi.(*WSClient).connectLocked()
      internal/redisapi/websocket.go:190
Previous write at 0x00c00018a2b0 by goroutine 15:
  ...redisapi.(*WSClient).reconnect()
      internal/redisapi/websocket.go:338
```

Corroborating evidence that this was already known to fire: `newTestWSClient` in
`websocket_write_test.go` disables reconnect with a comment saying reconnect's
unsynchronized mutation "would make `-race` report something unrelated to the
write path under test". That comment is updated here, since it now describes a
fixed bug and this repo has a documented history of stale comments costing time.

## Changes

- `internal/redisapi/websocket.go`
  - `nextReconnectDelay()` does the read, the doubling and the clamp in one
    critical section under `connMu`, and returns the delay. `connMu` is not held
    across the sleep.
  - `reconnect` sleeps on the returned value instead of re-reading the field.
    It used to read the field twice, once to log and once to sleep, so a reset
    landing between them made the node retry sooner than it had just announced.
  - `initialReconnectBackoff` replaces the two `time.Second` literals so the
    constructor and the reset in `connectLocked` cannot drift.
  - The field comment now names its guard.
- `internal/redisapi/websocket_backoff_test.go` (new), three tests.
- `internal/redisapi/websocket_write_test.go`, corrected the stale comment.

`connMu` is reused rather than adding a third mutex. The reset has to live in
`connectLocked`, whose caller already holds `connMu`, so this costs no new lock
and no nested acquire in a file whose lock discipline is already load-bearing.

## What this makes worse

- It widens `connMu`'s remit past "guards the conn pointer", which is how the
  existing comment describes it. A future reader has one more field to keep in
  mind, and `reconnect` now takes `connMu` twice per iteration (once shared for
  the connected check, once exclusive to advance the schedule) where it used to
  take it once.
- It adds a blocking point that did not exist. `Connect` holds `connMu` across
  the whole dial, bounded only by the 10s `HandshakeTimeout`, so
  `nextReconnectDelay` can now stall behind an in-flight dial where the old
  unlocked advance never blocked. It is benign in both directions that matter:
  the effect is a longer wait, never a shorter one, and `reconnect` already
  blocked the same way at its own `connMu.Lock()` a few lines later. It is still
  a new place the reconnect goroutine can park.
- Capturing the delay in a local is a behaviour change, not just a lock. Today a
  `Connect()` that succeeds between the log line and the sleep shortens that one
  sleep. After this it does not. The new behaviour is the more conservative of
  the two and matches what was logged, but it is a change.
- The advance now happens before the attempt rather than after it, so during a
  sleep the field holds the *next* delay rather than the current one. The
  sequence of sleeps is unchanged and nothing else reads the field, so this only
  matters to a future debugger.
- Most importantly: `reconnect` now looks fully synchronized while two real bugs
  in the same function are untouched, so it invites over-trust. Both are filed
  rather than fixed here, because #725 and #723 were deliberately sequenced one
  concern at a time and this is the third pass over the same lines:
  - #740, a `Connect()` during a reconnect starts a **second** `readLoop` on the
    same gorilla `Conn`. gorilla permits one concurrent reader just as it permits
    one concurrent writer, so this is the read-side twin of #720. It was observed
    under `-race` while building the test here, which is why the racing goroutine
    in that test calls `connectLocked` under `connMu` rather than `Connect`.
  - #741, `reconnect` checks `c.done` at the top of the loop but not after the
    sleep, so `Close` during a backoff still lets it dial and mark the client
    connected afterwards.

## Testing

`cpuslot go test -race ./internal/redisapi/...` passes, run five times. Full
`go test ./...` and `go vet ./...` pass.

No test asserts on wall-clock elapsed time. The schedule tests assert on returned
values; the race test asserts nothing and relies on the detector.

Mutation results, one deliberate break per test:

| Mutation | Effect |
| --- | --- |
| Drop the `connMu` acquire from `nextReconnectDelay` (the pre-fix shape) | `TestReconnectBackoffAccessIsSynchronized` FAILS under `-race`, and `TestNextReconnectDelayAdvancesAtomically` FAILS |
| Remove the `maxBackoff` clamp | `TestNextReconnectDelayAdvancesAndClamps` FAILS |
| Split the accessor into a locked load and a separate locked store | `TestNextReconnectDelayAdvancesAtomically` FAILS (duplicate delays handed out, so a doubling was lost) |
| Remove the reset in `connectLocked` | `TestSuccessfulConnectResetsReconnectBackoff` FAILS with `delay after a successful connect = 42s, want 1s` |

The split load/store mutation initially survived a single-round version of the
atomicity test, so that test now repeats the trial 300 times and kills it on
every run.

### DIFF
--- a/internal/redisapi/websocket.go
+++ b/internal/redisapi/websocket.go
@@ -54,7 +54,17 @@ type WSClient struct {
 	done     chan struct{}
 	stopOnce sync.Once
 
-	// Reconnection settings
+	// Reconnection settings.
+	//
+	// reconnectBackoff is guarded by connMu (#728). It is written by
+	// connectLocked, whose caller already holds connMu for write, and it is
+	// advanced by nextReconnectDelay, which takes connMu for the whole
+	// read-modify-write and releases it before the caller sleeps. Nothing may
+	// touch it unlocked: reconnect used to, and a torn or stale read there
+	// shortens a backoff.
+	//
+	// reconnectEnabled and maxBackoff are set once in NewWSClient and never
+	// written again, so they need no guard.
 	reconnectEnabled bool
 	reconnectBackoff time.Duration
 	maxBackoff       time.Duration
@@ -102,6 +112,11 @@ type WSMessage struct {
 	Data      map[string]string `json:"data,omitempty"`      // Job data fields from stream
 }
 
+// initialReconnectBackoff is the delay before the first reconnect attempt, and
+// the value the schedule returns to after any successful connect. It lives in
+// one place so the constructor and the reset in connectLocked cannot drift.
+const initialReconnectBackoff = time.Second
+
 // NewWSClient creates a new WebSocket client.
 func NewWSClient(cfg WSClientConfig) *WSClient {
 	reconnectEnabled := cfg.ReconnectEnabled
@@ -117,7 +132,7 @@ func NewWSClient(cfg WSClientConfig) *WSClient {
 		subscriptions:    make(map[string]bool),
 		done:             make(chan struct{}),
 		reconnectEnabled: reconnectEnabled,
-		reconnectBackoff: time.Second,
+		reconnectBackoff: initialReconnectBackoff,
 		maxBackoff:       time.Minute,
 		writeTimeout:     defaultWriteTimeout,
 		debugFunc:        cfg.DebugFunc,
@@ -187,7 +202,10 @@ func (c *WSClient) connectLocked(ctx context.Context) error {
 
 	c.conn = conn
 	c.connected = true
-	c.reconnectBackoff = time.Second // Reset backoff on successful connection
+	// Reset the backoff schedule on a successful connection. The caller holds
+	// connMu for write, which is what makes this assignment safe against
+	// nextReconnectDelay running in a reconnect goroutine (#728).
+	c.reconnectBackoff = initialReconnectBackoff
 
 	c.debug("ws: connected successfully")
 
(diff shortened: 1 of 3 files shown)


sunapi386:
Superseded by #749, which merges this together with the other half and adds the reconnect jitter, the #741 fix and a combined regression test.

These two cannot ship independently: #745 makes `handleDisconnect` fire on every read-deadline expiry, which turns #728's unsynchronized `reconnectBackoff` from latent into routine. Suggest closing this in favour of #749 rather than merging both, since both restructure the same function and conflict.

sunapi386:
Superseded by #749, which integrates this with the other websocket.go change and adds reconnect jitter. Merged as e23cb6d.

Issue #722, opened 2026-08-10
A pub/sub publish failure aborts publishStatus before the durable node:status:stream write


sunapi386:
## What is broken

A failure on the real-time pub/sub channel also destroys the durable heartbeat,
because `publishStatus` returns before it gets to the stream write.

`internal/heartbeat/api.go`:

```go
// 1. Publish to Pub/Sub for real-time UI updates
if err := p.client.Publish(ctx, p.pubSubChannel, msg); err != nil {
    return fmt.Errorf("failed to publish to pub/sub: %w", err)
}
...
// 2. Add to Stream for reliable processing by Python worker
```

Step 2 (`XADD node:status:stream`) is the reliable path, and it is the one the
platform consumes to update the node's last-reported timestamp. Step 1 is the
best-effort path that drives live UI updates. The ordering makes the
best-effort path a hard gate on the reliable one, which inverts their intended
roles: the comments in this very function label them "real-time UI updates" and
"reliable processing".

## Consequence

Observed on a node where the pub/sub publish was failing for an unrelated
reason: the worker was healthy and consuming jobs (`consuming: true`,
`last_consume_status: 200`, polling every few seconds), but its status record
on the platform froze at the moment of the last successful publish and stayed
frozen for 12+ hours.

Downstream, the platform's fail-closed per-node routing refuses targeted
dispatch to a node with a stale heartbeat, so the node is removed from the
fabric for anything node-pinned despite being fully functional. That turns a
degraded telemetry channel into a capacity outage.

## Suggested fix

Attempt both publishes and only return an error if the durable one fails.
Roughly:

```go
pubErr := p.client.Publish(ctx, p.pubSubChannel, msg)
if pubErr != nil {
    p.debug("heartbeat: pub/sub publish failed (non-fatal): %v", pubErr)
}
// ... marshal payload ...
if err := p.client.StreamAdd(ctx, p.streamName, streamFields, maxLen); err != nil {
    return fmt.Errorf("failed to add to stream: %w", err)
}
return nil
```

so a pub/sub outage degrades live UI freshness only. Log the pub/sub error at a
lower level (or rate-limit it) so it does not spam a warning every interval,
which is what it does today.

Worth considering alongside: the node keeps publishing into a channel that has
been failing every 30s for hours without escalating. Some form of "my heartbeat
has not landed in N intervals" signal, surfaced in `citadel status`, would make
this self-evident on the node instead of only visible from the platform side.

## Related

[...]

PR #745, opened 2026-08-11
fix(redisapi): websocket ping/pong keepalive and read deadline (#734)

### LINKED ISSUE #734: No WebSocket ping/pong keepalive or read deadline: a half-open connection reports as healthy
## What is broken

The pub/sub WebSocket client has no ping/pong keepalive and no read deadline, so
a half-open connection is indistinguishable from a healthy one.

`readLoop` blocks in `ReadMessage` forever. If the peer goes away without a TCP
FIN that reaches us (a NAT or load balancer dropping idle state, a network
partition, a Railway instance replaced mid-flight), no read error is ever
delivered. `connected` stays `true`.

## Consequence

- `Publish` sees `IsConnected() == true` and sends over the dead socket. Writes
  into a half-open connection succeed locally, so the publish **appears to
  succeed and silently goes nowhere**.
- `handleDisconnect` is never called, so reconnect never fires.
- Once #723 lands, `citadel status` will report the pub/sub transport as
  `websocket` while nothing is flowing, so the new observability line is
  confidently wrong in exactly this case. This bounds what that PR can claim.

This is the same failure shape as the incident that produced #720 through #723:
telemetry that is broken while reporting itself as fine. Heartbeats would stop
landing, the platform would mark the node offline after 120s, and per #722's
chain the node would be refused for targeted dispatch, all while the node
believes its WebSocket is up.

## Suggested fix

Standard gorilla keepalive discipline:

- `SetReadDeadline` on the connection, extended by a pong handler.
- A ping ticker at some fraction of the read deadline, writing with
  `WriteControl` (documented safe concurrently with other methods, so it does
  not need the write mutex added in #725).
- On read deadline expiry, treat it as a disconnect: `handleDisconnect` and
  reconnect.

Pick the intervals against the heartbeat cadence (30s) so a dead socket is
detected well inside the platform's 120s offline sweep, otherwise the node still
goes offline before it notices.

Once this exists, the `citadel status` pub/sub line from #723 becomes
trustworthy rather than merely present.

## Found by

Raised while implementing #723; deliberately not folded into that PR, which is
already changing connect and retry behaviour.

## Related

- #723 one-shot connect retry, and the observability line this bounds
- #725 write mutex and write deadline
- #728 reconnectBackoff data race
- #722 the offline chain this would trigger

sunapi386:
## Context

Closes #734.

The pub/sub WebSocket client had no keepalive and no read deadline, so `readLoop`
blocked in `ReadMessage` forever. A peer that went away without a FIN reaching us
(NAT or LB dropping idle state, a partition, a Railway instance replaced
mid-flight) delivered no read error, so `connected` stayed `true`, `Publish`
wrote into a dead socket and reported success, and `handleDisconnect` never
fired, so reconnect never happened.

Node 1297 now reports `pubsub_transport: websocket` in `citadel status`
(v2.103.0). Without this, that line can read `websocket` while nothing is
flowing, which is the same shape as the twelve-hour outage behind #720-#723:
telemetry broken while reporting itself fine.

## The real payoff: an honest `IsConnected()`

`Client.Publish` already branches on `wsClient.IsConnected()` and falls back to
HTTP when it is false. The fallback exists; what was missing was any way for
`IsConnected()` to become false. That makes the two error directions very
asymmetric:

- **False positive** (tear down a live connection): publishes degrade to HTTP for
  ~1s and the client redials. Cheap, bounded, self-correcting.
- **False negative** (miss a dead connection): `connected` stays true, the
  fallback never engages, and every publish is silently lost forever.

That asymmetry is what licenses erring toward detection. This PR is less "detect
inside 120s" than "make `IsConnected()` honest so the fallback that already
exists can engage".

## Changes

- `armKeepalive` sets a read deadline on a freshly dialed connection and installs
  pong and ping handlers that push it out. Called from `connectLocked`, so it
  covers reconnects, not just the first `Connect`.
- `readLoop` extends the deadline after every successful read, so ordinary
  application traffic counts as liveness and not only pongs, and reports a
  deadline expiry distinctly from an ordinary read error.
- `pingLoop` sends a ping every `pingInterval`. It is long-lived and re-reads
  `c.conn` each tick, so it spans reconnects.
- A read timeout routes through the existing `handleDisconnect` -> `reconnect`
  path; no new teardown mechanism.
- The pong we send in answer to a server ping is bounded at one second, matching
  gorilla's default handler. That reply runs on the read goroutine and
  `WriteControl` waits on gorilla's write semaphore, which an in-flight
  `WriteMessage` can hold for `defaultWriteTimeout`, so a longer bound there
  would park the only reader behind write congestion. Skipping a pong is
  cheaper: the timeout is already swallowed, so we keep reading.
- `loopsOnce` guards both background goroutines. They were already designed to
  outlive a single connection, but `Connect` is callable again whenever
  `connected` is false (#723's background retry loop does exactly that), and a
  second `Connect` would have started a second reader. gorilla permits one
  concurrent reader.

## Intervals: 15s ping, 45s read deadline

Sized against the platform's 120s offline sweep rather than network latency. If
detection is slower than the sweep the node is already marked offline (and
refused for targeted dispatch, #722) by the time it notices, which defeats the
point. 45s detection plus ~1s reconnect backoff leaves better than 2x margin and
lands inside two 30s heartbeat intervals.

The ping is one third of the deadline on purpose, so three pings go unanswered
before we act. That absorbs a dropped pong and scheduler jitter on a
GPU-saturated node. Tightening further buys little (the sweep, not us, sets the
deadline that matters) and costs false-positive teardowns.

## Why the ping does not take `writeMu`

Verified against the vendored `gorilla/websocket` source rather than taken on
faith, since #720 was a panic from exactly this class of mistake and #725's mutex
is now load-bearing. `WriteControl`:

- takes gorilla's own single-writer semaphore `c.mu` for the duration of the
  frame, so it cannot interleave with a `WriteMessage` frame,
- reads and latches `writeErr` under `writeErrMu`,
- calls `c.conn.SetWriteDeadline` on the underlying `net.Conn` (goroutine-safe)
  instead of touching the `c.writeDeadline` field that `writeMu` exists to guard.

`doc.go` says the same in one line: "The Close and WriteControl methods can be
called concurrently with all other methods." There is a test that would catch a
regression here rather than relying on the reading (see the mutation table).

## What this makes WORSE

1. **#728 becomes substantially more reachable.** Before this change
   `handleDisconnect` almost never fired, which was the bug. Now it fires on
   every deadline expiry, so `go c.reconnect()` runs far more often.
   `reconnect()` does `c.reconnectBackoff *= 2` with no `connMu` while
   `connectLocked` writes `c.reconnectBackoff = time.Second` under
   `connMu.Lock`. Not fixed here to keep this PR to one subject, but this PR is
   what makes that race worth prioritising. Note the new tests do not prove it is
   absent: in them both writes happen on the same goroutine, so `-race` is clean
   for the wrong reason.
2. **A new false-positive class, and it flaps rather than fails.** A peer that is
   alive but never answers a ping (a proxy that strips control frames, a server
   that stops ponging while otherwise fine) is now torn down every 45s. Because
   `connectLocked` resets the backoff to 1s on every successful dial, that is a
   redial roughly every 46s per node, forever, against an endpoint that returns
   429 (typed by #723) and fleet-synchronized with no jitter. Previously those
   nodes sat quiet. Mitigated by counting any inbound frame as liveness, and the
   fabric route is served by `ws`, which answers pings automatically, but the
   failure mode is real and belongs on the record.
3. **The ping is a real writer.** Even with `WriteControl`, a ping can block up
   to its own 10s deadline behind a large in-flight `WriteMessage`, and a hard
   failure inside `WriteControl` latches `writeErr` for the whole connection, so
   a ping can kill a connection that would otherwise have kept serving reads.
   This is why a failed ping is advisory (logged, no teardown): the read deadline
   is the single authority on liveness, a queued ping must not be able to cause
   an outage, and a genuinely fatal failure is not lost (the next publish tears
   down, and no pongs arrive either way).
4. **One control frame per node per 15s**, four times the heartbeat rate on the
   fabric ws endpoint. Tiny per node, nonzero at fleet scale.
5. **The read deadline bounds a single inbound message, not just idle time.**
   Frames on this route are small JSON, so this is theoretical, but a read taking
   more than 45s would now be cut.

The one diagnostic that matters here (`ws: no traffic for 45s, treating
connection as dead`) does reach the log on the real path: `cmd/work.go` passes
`DebugFunc: Debug` into the API source, which passes it to `redisapi.NewClient`,
and `Debug` writes to `latest.log` unconditionally. `internal/chat` does not pass
one, so a chat-client flap stays silent there.

## Testing

`cpuslot go test -race ./internal/redisapi/...` is clean, as is `go vet ./...`
and the full `go test ./...`. (`TestNullSinkRecorder_CapturesAudio` in
`internal/platform` flaked once under a parallel run and passes repeatedly on
its own; it captures from PulseAudio and is unrelated to this package.)

The keepalive tests run at a 250ms ping and a 1s read deadline: production's 1:4
ratio, but coarse on purpose. CI runs on a shared, contended runner under
`-race`, and a sub-second budget would turn an ordinary scheduling hiccup into a
red build that gets rerun rather than diagnosed.

Every new test was mutation-tested: the fix was deliberately broken and each test
confirmed to fail. Per test:

| Mutation | Caught by |
|---|---|
| M1: drop the initial `SetReadDeadline` in `armKeepalive` | `TestReadDeadlineTearsDownASilentConnection` (uniquely), `TestReadDeadlineDisconnectTriggersReconnect`, `TestKeepaliveIsRearmedOnEveryReconnect` |
| M2: pong handler returns nil instead of extending the deadline | `TestKeepaliveHoldsAQuietButHealthyConnection`, `TestReadDeadlineDisconnectTriggersReconnect` |
| M3: never start `pingLoop` | `TestPingsReachTheServer` (uniquely), `TestKeepaliveHoldsAQuietButHealthyConnection`, `TestReadDeadlineDisconnectTriggersReconnect` |
| M4: ping via `SetWriteDeadline` + `WriteMessage` instead of `WriteControl` | `TestPingsAreSafeAlongsideConcurrentWrites` (uniquely; panics `concurrent write to websocket connection` from `flushFrame`, which is #720) |
| M5: `handleDisconnect` never schedules a reconnect | `TestReadDeadlineDisconnectTriggersReconnect` (uniquely) |
| M6: arm the keepalive in `Connect` instead of `connectLocked` | `TestKeepaliveIsRearmedOnEveryReconnect` (uniquely) |

The trailing flap check in `TestReadDeadlineDisconnectTriggersReconnect` (stay
connected for `3 * pongWait` after recovery, and no further dials) catches M2 and
M3 but nothing uniquely; it is a guard, not a demonstrated tripwire.

M6 is worth calling out: the obvious assertion (reconnect, then check the new
socket stays up) does NOT catch it, because an unarmed connection is more stable,
not less. Nothing can declare it dead. The test had to be rewritten to face a
permanently silent server and assert the client keeps detecting and redialing.

Not covered by an automated test, and deliberately so: whether Railway's proxy
and the `ws`-based fabric route pass control frames end to end in production.
That is what worse-item 2 hangs on. Worth watching node 1297's log for
`ws: no traffic for` after this ships; a steady 46s cadence there means the
premise is wrong and the intervals need revisiting.

### DIFF
--- a/internal/redisapi/websocket.go
+++ b/internal/redisapi/websocket.go
@@ -3,7 +3,9 @@ package redisapi
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -42,6 +44,13 @@ type WSClient struct {
 	// Overridden in tests; see defaultWriteTimeout for the value rationale.
 	writeTimeout time.Duration
 
+	// pingInterval and pongWait drive the keepalive (#734). pongWait is the
+	// read deadline: nothing arriving within it means the socket is dead.
+	// pingInterval is how often we send a ping to provoke traffic. Both are
+	// overridden in tests; see defaultPongWait for the value rationale.
+	pingInterval time.Duration
+	pongWait     time.Duration
+
 	// Message handlers
 	handlers   map[string]func(WSMessage)
 	handlersMu sync.RWMutex
@@ -53,6 +62,14 @@ type WSClient struct {
 	// Control channels
 	done     chan struct{}
 	stopOnce sync.Once
+
+	// loopsOnce guards the background goroutines. readLoop and pingLoop are
+	// deliberately long-lived: they re-read c.conn every iteration, so they
+	// survive a reconnect and must NOT be started per connection. Connect is
+	// callable again whenever connected is false (the #723 retry loop does
+	// exactly that), and without this a second Connect would start a second
+	// reader -- gorilla permits exactly one concurrent reader.
+	loopsOnce sync.Once
 
 	// Reconnection settings
 	reconnectEnabled bool
@@ -120,6 +137,8 @@ func NewWSClient(cfg WSClientConfig) *WSClient {
 		reconnectBackoff: time.Second,
 		maxBackoff:       time.Minute,
 		writeTimeout:     defaultWriteTimeout,
+		pingInterval:     defaultPingInterval,
+		pongWait:         defaultPongWait,
 		debugFunc:        cfg.DebugFunc,
 	}
 }
@@ -144,8 +163,11 @@ func (c *WSClient) Connect(ctx context.Context) error {
 		return err
 	}
 
-	// Start read loop
-	go c.readLoop()
+	// Both loops outlive any single connection; see loopsOnce.
+	c.loopsOnce.Do(func() {
+		go c.readLoop()
+		go c.pingLoop()
+	})
 
 	return nil
 }
@@ -184,6 +206,15 @@ func (c *WSClient) connectLocked(ctx context.Context) error {
 		}
 		return fmt.Errorf("WebSocket connection failed: %w", err)
 	}
+
+	// Arm the keepalive BEFORE publishing the connection (#734).
+	//
+	// SetReadDeadline, SetPongHandler and SetPingHandler are gorilla READ
+	// methods, so they must not run concurrently with ReadMessage. Doing this
+	// while conn is still private -- readLoop cannot see it until c.conn is
+	// assigned under connMu -- is what makes that safe. The handlers only ever
+	// fire from inside ReadMessage afterwards, i.e. on the read goroutine.
+	c.armKeepalive(conn)
 
 	c.conn = conn
 	c.connected = true
@@ -254,10 +285,25 @@ func (c *WSClient) readLoop() {
 
 		_, message, err := conn.ReadMessage()
 		if err != nil {
-			c.debug("ws: read error: %v", err)
+			var netErr net.Error
+			if errors.As(err, &netErr) && netErr.Timeout() {
+				// The read deadline expired: nothing at all arrived for
+				// pongWait, not even a pong for our pings. Before #734 this
+				// case did not exist and the loop simply blocked forever, so a
+				// half-open socket stayed "connected" while nothing flowed.
+				c.debug("ws: no traffic for %v, treating connection as dead", c.pongWait)
+			} else {
+				c.debug("ws: read error: %v", err)
+			}
 			c.handleDisconnect(conn)
 			continue
 		}
+
+		// Any inbound frame is evidence the peer is alive, so extend the
+		// deadline on data too, not just on pongs. That matters if a proxy or
+		// server ever stops honoring ping frames: a busy connection then still
+		// stays up on its own traffic.
+		_ = conn.SetReadDeadline(time.Now().Add(c.pongWait))
 
 		var msg WSMessage
 		if err := json.Unmarshal(message, &msg); err != nil {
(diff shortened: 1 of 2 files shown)


sunapi386:
Superseded by #749, which merges this together with the other half and adds the reconnect jitter, the #741 fix and a combined regression test.

These two cannot ship independently: #745 makes `handleDisconnect` fire on every read-deadline expiry, which turns #728's unsynchronized `reconnectBackoff` from latent into routine. Suggest closing this in favour of #749 rather than merging both, since both restructure the same function and conflict.

sunapi386:
Superseded by #749, which integrates this with the other websocket.go change and adds reconnect jitter. Merged as e23cb6d.

Issue #748, opened 2026-08-11
reconnect does not re-check connected after the backoff sleep, so a racing Connect leaks an orphaned connection


sunapi386:
## What is broken

`WSClient.reconnect` checks `c.connected` under `connMu.RLock` BEFORE its
backoff sleep, and never again. After the sleep it takes `connMu` for write and
calls `connectLocked` unconditionally:

```go
c.connMu.RLock()
alreadyConnected := c.connected
c.connMu.RUnlock()
if alreadyConnected {
    return
}
... sleep ...
c.connMu.Lock()
err := c.connectLocked(context.Background())   // no re-check
c.connMu.Unlock()
```

A `Connect()` that lands inside that sleep window succeeds and sets
`c.conn = A`, `connected = true`. The reconnect goroutine then wakes and dials
again, and `connectLocked` overwrites `c.conn = B` without closing A.

This is the sibling of #741 (which is about `c.done` not being re-checked after
the same sleep, and is fixed on the integration branch); this one is about
`c.connected`.

## Consequence

Connection A is orphaned: still open at both ends, never read from (readLoop
follows `c.conn`, which is now B), never closed by anything. A file descriptor
and a server-side connection leak per occurrence.

The window is the whole backoff sleep, and it is entered whenever any
`EnableWebSocket` caller lands inside a reconnect (`enableWebSocketWithRetry`,
`worker.WSSource.Connect`, `chat.Client.Connect`, `controlcenter.go` — the same
overlap set as #740). #734's keepalive makes reconnect run on every read-deadline
expiry, so the number of windows per node per hour goes up sharply.

## Why it is not a one-line fix

The obvious fix (re-check `c.connected` after taking `connMu` and return) also
skips the reconnect callbacks that `reconnect` fires on success. Those callbacks
are how consumers re-arm after a reconnection (`OnReconnect`), so returning
early without firing them risks a client that is connected while nothing flows,
which is the exact failure shape #734 exists to prevent. Any fix has to decide
who fires the callbacks when `Connect` wins the race, and that decision belongs
in its own change rather than bolted onto an integration branch.

## Suggested fix

Make connection ownership explicit, per #740's option 1: a flag under `connMu`
that `reconnect` sets on entry and clears on return, so `Connect` can decline (or
wait) rather than racing. Then define the callback contract for the case where
`Connect` establishes the connection a reconnect was about to establish.

## Related

- #740 the same overlap, read-loop side (fixed by loopsOnce)
[...]

Issue #747, opened 2026-08-11
WSClient.reconnect discards the typed 429 retry_after that connectLocked builds


sunapi386:
## What is broken

`connectLocked` goes out of its way to type a rejected upgrade:

```go
if resp.StatusCode == http.StatusTooManyRequests {
    return fmt.Errorf("WebSocket connection failed: %w", rateLimitFromUpgrade(resp))
}
```

The comment above it says the point is so "callers that retry (issue #723) can
honor retry_after instead of falling into a generic backoff and polling tighter
than the server asked".

`WSClient.reconnect` is such a caller, and it discards it:

```go
if err != nil {
    c.debug("ws: reconnect failed: %v", err)
    continue
}
```

It then doubles its own schedule and tries again on its own timetable. Only
`cmd/work.go`'s `connectWithBackoffLabeled` actually honors the hint, and that
loop stops permanently on the first success, so after a node has connected once
the only remaining reconnector is the one that ignores it.

## Consequence

Against a rate-limited endpoint the client retries on a schedule the server did
not ask for. If `retry_after` is longer than the client's current step the node
polls tighter than requested, which is the #443 shape (a tight retry loop burned
the node's daily Redis-API quota and locked the node out for about a day).

This matters more after #734: `reconnect` used to run almost never and now runs
on every read-deadline expiry.

## Suggested fix

In `reconnect`, check `redisapi.AsRateLimitError(err)` on a failed dial and, when
present, sleep the server-requested interval instead of the local schedule (in
`done`-aware chunks, the way `sleepUntilCtx` does in `cmd/work.go`). Reset the
local schedule in that case, as `connectWithBackoffLabeled` does, since the
server is the authority.

Worth noting for prioritisation: for a 429ing endpoint this is worth more than
any client-side jitter guess, because it uses the number the server actually
sent instead of one we invented.

## Related

- #723 / PR #732, which established the honor-retry_after policy
- #734 keepalive, which makes this path hot
- #443

sunapi386:
Fixed in #750 (squash-merged to main). The `Closes` list only auto-closed #746; closing this manually. Regression test in `websocket_reconnect_cluster_test.go`, mutation-verified.

Issue #746, opened 2026-08-11
reconnect backoff resets on a connection that never proved itself, so a ping-less peer redials forever without backing off


sunapi386:
## What is broken

`connectLocked` resets `reconnectBackoff` to one second on EVERY successful
dial, with no regard for whether the connection it just established goes on to
be useful.

That is correct for the case it was written for (a blip, a redial, a connection
that then serves for hours). It is wrong for a connection that dies immediately
and repeatedly, which the keepalive in #734 makes an ordinary state rather than
an exotic one.

## Consequence

A peer that is alive but never answers a ping (a proxy that strips control
frames, a server that stops ponging while otherwise fine) is torn down every
`pongWait`. The dial that follows succeeds, so the schedule resets to one
second, so the next teardown is another `pongWait` away. The node redials
roughly every `pongWait + 1s`, forever, instead of backing off. #745's own
"what this makes worse" section describes exactly this.

Jitter on the reconnect delay does not fix it. The cycle's period is dominated
by `pongWait`, not by the backoff, so one second of jitter per 46-second cycle
random-walks the phases of a fleet apart far too slowly to matter. Jitter
addresses the mass-disconnect herd, which is a different failure.

## Suggested fix

Reset the schedule only for a connection that proved itself, e.g. record when
the connection was established and reset only once it has survived some minimum
(a few multiples of `pongWait`), rather than resetting at dial time. A
connection that never lasts then backs off toward `maxBackoff` like any other
repeated failure.

Deliberately NOT folded into the integration branch that raised it: this changes
what "a successful connect" means to the schedule, and that file already has a
history of partial fixes making things worse.

## Related

- #734 keepalive, which makes this reachable
- #728 the schedule's synchronization
- #443 the restart storm that makes retry rate load-bearing here

Issue #443, opened 2026-07-03
Worker crash-loops on Redis-API 429, self-DoSing into the daily rate limit (24h node lockout)


sunapi386:
## Incident
On node ubuntu-gpu the Citadel worker got stuck in a systemd restart storm and locked itself out of the Redis job API for ~24h. Journal (user `citadel.service`):

```
--- Services started ---
Error: Failed to connect to Redis API: ping failed with status 429:
{"error":"Rate limit exceeded","limit":50000,"window":"day","retry_after":86400,
 "reset_at":"2026-07-04T02:28:33Z"}
citadel.service: Main process exited, code=exited, status=1/FAILURE
citadel.service: Scheduled restart job, restart counter is at 5.
Started citadel.service ...
(repeats every ~10s)
```

## Root cause (crash-loop + no backoff = self-DoS)
The worker treats a failed Redis-API connect at startup as fatal: it `exit(1)`s. The systemd unit has `Restart=on-failure` with a short `RestartSec` (~10s), so it immediately relaunches, re-pings the SAME rate-limited API, gets 429 again, exits again. Each restart also re-runs service startup (docker service checks, etc.) and issues API calls. Over hours this restart storm consumed the node's **50,000 requests/day** Redis-API quota, so now every attempt returns 429 with `retry_after: 86400` and the node is hard-locked until `reset_at` (~24h). It cannot rejoin the queue in the meantime, and the loop keeps hammering a limit it can never satisfy.

Contributing factor: a stale duplicate worker (an old, since-deleted binary launched manually from a shell, running alongside the systemd unit) amplified the churn and request volume. Duplicate workers on one node also cause non-deterministic job routing (an AGENT_UPDATE landed on the stale old-version worker: `node 2.57.0 has no handler`).

## Fix directions
1. **Do not crash-loop on a transient/429 connection failure.** The worker should back off and retry IN-PROCESS with exponential backoff + jitter rather than `exit(1)` -> instant systemd restart. A failed control-plane connect is expected/recoverable, not fatal.
2. **Honor `retry_after` / `reset_at`.** On a 429 with `retry_after: 86400`, do not retry for 24h -- sleep toward the reset (capped) instead of hammering. Parse the structured 429 body.
[...]

PR #749, opened 2026-08-11
fix(redisapi): integrate the backoff-race and keepalive fixes, and jitter the reconnect (#728, #734, #740, #741)

### LINKED ISSUE #728: reconnectBackoff is written under connMu.Lock in one place and unlocked in another

### LINKED ISSUE #734: No WebSocket ping/pong keepalive or read deadline: a half-open connection reports as healthy

### LINKED ISSUE #740: Connect during a reconnect starts a second readLoop on the same gorilla Conn

### LINKED ISSUE #741: reconnect can dial after Close because done is not re-checked after the backoff sleep

sunapi386:
## Context

Integration branch for PR #743 (#728, `reconnectBackoff` synchronization) and PR
#745 (#734, ping/pong keepalive and read deadline), plus the jitter and the #741
fix that integrating them makes necessary. Opened as a draft for a human to
merge; #743 and #745 should be closed in favour of this rather than merged
alongside it, since both restructure the same function.

**They are not independent, and #745 must not ship without #743.** #745's own
author says so: before the keepalive, `handleDisconnect` almost never fired, so
`reconnect`'s unsynchronized `reconnectBackoff *= 2` almost never ran against
`connectLocked`'s `reconnectBackoff = time.Second` under `connMu`. After the
keepalive it fires on every read-deadline expiry. #745 turns a latent data race
into a routine one. Shipping it alone would be the third consecutive change to
this file that fixed one thing and made the next thing worse.

## How the conflict was resolved

One textual conflict, in the `WSClient` struct field block: both PRs inserted
above `// Reconnection settings`. Both sides kept, `loopsOnce` first.

Everything else auto-merged, and the two overlaps that matter were verified by
hand rather than trusted:

- **`connectLocked`.** #745 inserts `c.armKeepalive(conn)` immediately before
  `c.conn = conn`; #743 replaces the `c.reconnectBackoff = time.Second` line two
  lines later with `initialReconnectBackoff` plus a comment naming its guard.
  Both survive, in that order. The ordering is load-bearing on #745's side (the
  keepalive must be armed while `conn` is still private, because
  `SetReadDeadline` / `SetPongHandler` / `SetPingHandler` are gorilla READ
  methods and must not race `ReadMessage`), and on #743's side (the reset is
  safe because the caller holds `connMu`). Pinned by a mutation below.
- **`reconnect`'s sleep.** #745 does not touch it; #743 replaces the double read
  of the field with `nextReconnectDelay()`. #743's shape is kept, and it is the
  line the jitter and the #741 fix then modify.

No behaviour from either PR was dropped. Commits are ordered so a reviewer can
read the two merges separately from the new work.

## Jitter: additive and upward-only, `[base, 2*base)`

There was no randomness anywhere in this file. Every node ran exactly
1, 2, 4, ... 60 seconds, so a control-plane blip disconnects the whole fleet at
once and it redials in lockstep against an endpoint that answers 429. #734 makes
this worse rather than better, because the path now runs routinely.

The scheme is `delay = base + rand[0, base)`, where `base` is the deterministic
step. `nextReconnectDelay` returns the jittered value; `reconnectBackoff` keeps
storing the clean doubling, so nothing accumulates randomness and
`connectLocked`'s reset still means exactly "back to one second".

**Full jitter, `rand(0, base)`, was rejected** because its floor is zero: a node
just answered 429 could redial milliseconds later, and its mean delay is half the
current schedule, so fleet-wide retry pressure doubles. That is the #443 shape
(tight retries burned a node's daily Redis-API quota and locked it out for about
a day), and #728 explicitly names too-aggressive as the direction that matters.
`cmd/work.go` already refuses to "poll tighter than retry_after" for the same
reason; a client-side scheme that halves the mean would contradict it.

**Decorrelated jitter, `min(cap, rand(base, prev*3))`, was rejected** for a
weaker version of the same reason (it can return to `base` repeatedly, so a long
outage can sit near one second), and because its state is the previous SLEEP,
which collides with `reconnectBackoff` holding the deterministic schedule that
`connectLocked` resets.

Additive jitter is never shorter than what ships today, and its spread is the
full width of the current step, which is exactly what decorrelates a fleet whose
nodes all hold the same step after a shared blip.

**The jittered sleep is deliberately not clamped to `maxBackoff`.** `maxBackoff`
caps the schedule, not the sleep. Clamping the result would collapse the spread
to zero at the ceiling, which is precisely the worst case: a long outage puts
every node at the cap and they would all dial together again. This has its own
test and its own mutation.

### What the jitter does NOT fix

The failure #745 documented (a peer that is alive but never pongs, torn down
every `pongWait`, redialing at one second, forever, fleet-synchronized) is **not
fixed by jitter**, and the PR should not be read as claiming it is. That cycle's
period is dominated by `pongWait` (45s), not by the backoff (1s), so one second
of jitter per 46-second cycle random-walks the phases apart far too slowly to
matter at fleet scale. Jitter fixes the mass-disconnect herd, which is the
genuinely dangerous one against a rate-limited endpoint.

Two things would actually address the flap, both filed rather than done here:

- #746, the schedule resets on a connection that never proved itself.
- #747, `reconnect` discards the typed 429 `retry_after` that `connectLocked`
  goes out of its way to build. **For a 429ing endpoint this outranks the
  jitter**, because it uses the number the server actually sent instead of one we
  invented.

## #740: fixed, with a regression test. Mechanism.

`loopsOnce` makes a second reader structurally impossible, and the reason is
narrower than "one goroutine":

1. `readLoop` is started inside `c.loopsOnce.Do` in `Connect`, so at most one
   `readLoop` goroutine exists for the client's whole life.
2. It has exactly one `return`, the `<-c.done` case, so it cannot exit early and
   leave `loopsOnce` spent with no reader. (Verified: one `return` in the
   function.)
3. It is the only caller of `ReadMessage` in the package. `pingLoop` writes with
   `WriteControl`; the pong and ping handlers installed by `armKeepalive` run
   from inside `ReadMessage`, i.e. on the read goroutine.
4. The swap window does not create a second reader either, and this is the part
   the goroutine count alone does not cover. `Connect` dials only when
   `c.connected` is false, and `connected` goes false only in `handleDisconnect`
   (or `Close`), which under `connMu` closes the connection and nils `c.conn`
   BEFORE releasing the lock. So there is never a moment where `Connect` dials a
   second connection while `readLoop` is blocked on a live current one: the one
   it is blocked on has already been closed, gorilla documents `Close` as safe
   alongside any other method, so the blocked `ReadMessage` returns promptly,
   and `handleDisconnect`'s connection-pointer guard then makes that stale error
   a no-op against the new connection.

Also checked, because it is the "IsConnected() stays true and nothing flows"
shape: `reconnect -> connectLocked` never touches `loopsOnce`, so a client whose
first successful dial came via reconnect would be connected with no reader. That
is unreachable. `connectLocked` has exactly two callers (`Connect`, `reconnect`);
`reconnect` is spawned only from `handleDisconnect`; `handleDisconnect` is called
only from `readLoop` and `sendMessage`, and both require a prior successful
`Connect`. `Connect` runs `loopsOnce.Do` while still holding `connMu`, so the
loops are running before any other goroutine can observe `connected == true`.

That is the argument. The tripwire is mutation K below: restoring the
pre-`loopsOnce` bare `go c.readLoop()` in `Connect` makes
`TestConnectRacesKeepaliveTeardownOnTheBackoffSchedule` report the exact trace
#740 was filed with, two `readLoop` goroutines inside gorilla's `NextReader` on
the same `Conn`. So #740 is closed with coverage rather than with reasoning, and
a later refactor that drops `loopsOnce` cannot reintroduce it silently.

A comment with this mechanism has been posted on #740.

## #741: fixed here

Small, and squarely in scope: the keepalive changes its odds from rare to
routine (the window opens on every read-deadline expiry rather than almost
never), and the jitter lengthens the sleep the window lives in, up to two
minutes. `time.Sleep` is replaced by `sleepUnlessDone`, a select on a timer and
`c.done`, so `Close` during a backoff neither waits out the backoff nor lets a
dial happen afterwards.

The `context.Background()` sub-item of #741 is deliberately left: giving the
client a real context is a larger change than this loop, and it is noted in the
code.

## What the integrated result makes WORSE

1. **`connMu` contention on the reconnect path goes from exotic to routine.**
   #743 flagged `nextReconnectDelay` as "a new place the reconnect goroutine can
   park", because `Connect` holds `connMu` across a whole dial (bounded only by
   the 10s `HandshakeTimeout`). Integration is what makes that frequent: this
   path now runs on every read-deadline expiry per node. It is benign in
   direction (the effect is a longer wait, never a shorter one) but it is real.
2. **The worst-case pub/sub gap doubles, from one minute to two.** That is the
   price of keeping the spread alive at the ceiling. It is reached only after
   roughly seven consecutive failed dials, i.e. when the control plane is
   already down, and `Client.Publish` falls back to HTTP throughout. Stated
   plainly because it is a real regression in a number someone may be relying on.
3. **The next dial is no longer predictable from the log line.** `ws: attempting
   reconnect in %v` still reports the exact sleep, but two nodes at the same step
   no longer wait the same time, so "all nodes retry at t+4s" stops being a thing
   you can assume when reading logs across the fleet.
4. **The flap herd remains** (see above), and this PR adds the jitter that a
   future reader might mistake for a fix for it. That is why it is written out
   here and in the code, and why #746 and #747 exist.
5. **An orphaned-connection leak becomes more frequent, and is not fixed**
   (#748). `reconnect` re-checks `c.done` after the sleep now, but not
   `c.connected`, so a `Connect` that wins the race leaves the reconnect
   goroutine to dial a second connection that overwrites `c.conn` without
   closing the first. The keepalive multiplies the windows. It is not fixed here
   because the obvious one-line re-check would skip the `OnReconnect` callbacks
   that consumers use to re-arm, which is the "connected but nothing flows"
   failure this whole area exists to prevent. That decision deserves its own
   change.
6. **`reconnect` now looks even more finished than it is.** #743 made this point
   about itself; it is more true now. Three real defects remain in that function
   (#746, #747, #748) behind a body that reads as fully synchronized,
   cancellable and jittered.

## Testing

```
cpuslot go build ./... && cpuslot go vet ./...          clean
cpuslot go test ./...                                    all pass
cpuslot go test -race ./internal/redisapi/... ./internal/heartbeat/...
                                                         clean, run 3 times
```

No test asserts on elapsed wall-clock time. The jitter tests assert on the value
`nextReconnectDelay` returns with the randomness injected through a
`jitterFrac` field; the `Close` test asserts on a dial COUNT under a budget five
times the backoff. This is the failure mode `backoffSleep` in `cmd/work.go`
already documents: a timing assertion on a contended runner is flaky in the
direction that HIDES a regression, because an overloaded machine overshoots a
short sleep enough to make a collapsed backoff look correct.

`jitterFrac` is a field rather than a package var on purpose: a package var
mutated by one test can be read concurrently by a previous test's still-running
reconnect goroutine, which is a `-race` failure of the test harness's own making.

### Mutation results, per test

Every new or changed test was mutation-tested by breaking the specific thing it
covers and confirming that test fails. "uniquely" means no other test in the
package caught it.

| Mutation | Tests that FAIL |
|---|---|
| A: drop the `connMu` acquire from `nextReconnectDelay` (the pre-#743 shape) | `TestConnectRacesKeepaliveTeardownOnTheBackoffSchedule` (DATA RACE under `-race`), `TestReconnectBackoffAccessIsSynchronized`, `TestNextReconnectDelayAdvancesAtomically` |
| B: full jitter, `rand(0, base)` instead of `base + rand(0, base)` | `TestReconnectDelayIsNeverShorterThanTheSchedule` (both subtests), plus `TestReconnectDelaySpreadSurvivesAtTheMaximumBackoff`, `TestNextReconnectDelayAdvancesAndClamps`, `TestSuccessfulConnectResetsReconnectBackoff`, `TestCloseDuringTheBackoffPreventsAFurtherDial` |
| C: clamp the jittered delay at `maxBackoff` | `TestReconnectDelaySpreadSurvivesAtTheMaximumBackoff` (uniquely) |
| D: store the jittered value back into `reconnectBackoff` | `TestJitterDoesNotContaminateTheStoredSchedule` (uniquely among the schedule tests), `TestReconnectDelayIsNeverShorterThanTheSchedule/highest_draw` |
| E: revert `sleepUnlessDone` to a bare `time.Sleep` (the pre-#741 shape) | `TestCloseDuringTheBackoffPreventsAFurtherDial` (uniquely) |
| F: remove the `maxBackoff` clamp from the schedule | `TestNextReconnectDelayAdvancesAndClamps`, plus all three jitter tests |
| G: remove the backoff reset in `connectLocked` | `TestSuccessfulConnectResetsReconnectBackoff` (uniquely) |
| H: split `nextReconnectDelay` into a locked load and a separate locked store | `TestNextReconnectDelayAdvancesAtomically` (uniquely) |
| I: arm the keepalive in `Connect` instead of `connectLocked` (#745's M6) | `TestKeepaliveIsRearmedOnEveryReconnect` (uniquely) |
| K: remove `loopsOnce`, restore the bare `go c.readLoop()` in `Connect` (the pre-#745 shape, i.e. #740) | `TestConnectRacesKeepaliveTeardownOnTheBackoffSchedule` (uniquely; DATA RACE with two `readLoop` goroutines in `NextReader`, matching the trace in #740's body) |
| J: never start `pingLoop` (#745's M3) | `TestPingsReachTheServer`, `TestKeepaliveHoldsAQuietButHealthyConnection`, `TestReadDeadlineDisconnectTriggersReconnect` |

I and J are merge-integrity checks rather than new coverage: they target the
`connectLocked` and `Connect` regions the merge touched, and confirm #745's own
tripwires still fire after the conflict resolution. K is what earns the
"closes #740" above.

Mutation A is the one that justifies the new test. #743's race test deliberately
called `connectLocked` under `connMu` instead of the exported `Connect`, because
on that branch a `Connect` inside a reconnect window would trip #740 and fail
for the wrong reason. #745's tests put both writes on the SAME goroutine, so
`-race` was clean for the wrong reason there too. With `loopsOnce` in, the new
test drives the real path: every connection the server accepts goes silent, so
every one dies at the read deadline (`handleDisconnect -> go reconnect() ->
nextReconnectDelay`), while a background loop calls the exported `Connect`
whenever `IsConnected()` is false, which is the exact shape of
`EnableWebSocket` under #723's retry. Two writes, two goroutines, nothing
ordering them.

Not covered by an automated test, and deliberately: whether Railway's proxy and
the `ws` fabric route pass control frames end to end in production. That is what
#746 hangs on. Watch node 1297's log for `ws: no traffic for`; a steady cadence
at roughly `pongWait` means the premise is wrong and the intervals need
revisiting.

## Related

Closes #728, closes #734, closes #740, closes #741.
Follow-ups filed: #746, #747, #748.
Supersedes #743 and #745.

### DIFF
--- a/cmd/ws_pubsub.go
+++ b/cmd/ws_pubsub.go
@@ -62,8 +62,14 @@ func (w wsConnector) Connect(ctx context.Context) error {
 //
 // Re-arming after a LOST connection is already handled one layer down by
 // WSClient.handleDisconnect -> reconnect, which retries indefinitely and
-// restores subscriptions. This function deliberately does not duplicate that:
-// a second connect path could start a second readLoop on the same connection.
+// restores subscriptions. This function deliberately does not duplicate that.
+//
+// It used to be unsafe as well as redundant: a second connect path could start a
+// second readLoop on the same gorilla Conn, which permits exactly one concurrent
+// reader (#740). That is fixed (WSClient.loopsOnce starts the loops once for the
+// client's whole life), so the reason to keep a single re-arm path is now purely
+// that two of them would dial twice and leave one of the connections orphaned
+// (#748), not that they would corrupt the reader.
 // The returned channel closes when the retry loop has finished (immediately
 // when no loop was needed). Production callers ignore it; it exists so a test
 // can join the goroutine instead of sleeping and guessing.
--- a/internal/redisapi/websocket.go
+++ b/internal/redisapi/websocket.go
@@ -54,7 +54,17 @@ type WSClient struct {
 	done     chan struct{}
 	stopOnce sync.Once
 
-	// Reconnection settings
+	// Reconnection settings.
+	//
+	// reconnectBackoff is guarded by connMu (#728). It is written by
+	// connectLocked, whose caller already holds connMu for write, and it is
+	// advanced by nextReconnectDelay, which takes connMu for the whole
+	// read-modify-write and releases it before the caller sleeps. Nothing may
+	// touch it unlocked: reconnect used to, and a torn or stale read there
+	// shortens a backoff.
+	//
+	// reconnectEnabled and maxBackoff are set once in NewWSClient and never
+	// written again, so they need no guard.
 	reconnectEnabled bool
 	reconnectBackoff time.Duration
 	maxBackoff       time.Duration
@@ -102,6 +112,11 @@ type WSMessage struct {
 	Data      map[string]string `json:"data,omitempty"`      // Job data fields from stream
 }
 
+// initialReconnectBackoff is the delay before the first reconnect attempt, and
+// the value the schedule returns to after any successful connect. It lives in
+// one place so the constructor and the reset in connectLocked cannot drift.
+const initialReconnectBackoff = time.Second
+
 // NewWSClient creates a new WebSocket client.
 func NewWSClient(cfg WSClientConfig) *WSClient {
 	reconnectEnabled := cfg.ReconnectEnabled
@@ -117,7 +132,7 @@ func NewWSClient(cfg WSClientConfig) *WSClient {
 		subscriptions:    make(map[string]bool),
 		done:             make(chan struct{}),
 		reconnectEnabled: reconnectEnabled,
-		reconnectBackoff: time.Second,
+		reconnectBackoff: initialReconnectBackoff,
 		maxBackoff:       time.Minute,
 		writeTimeout:     defaultWriteTimeout,
 		debugFunc:        cfg.DebugFunc,
@@ -187,7 +202,10 @@ func (c *WSClient) connectLocked(ctx context.Context) error {
 
 	c.conn = conn
 	c.connected = true
-	c.reconnectBackoff = time.Second // Reset backoff on successful connection
+	// Reset the backoff schedule on a successful connection. The caller holds
+	// connMu for write, which is what makes this assignment safe against
+	// nextReconnectDelay running in a reconnect goroutine (#728).
+	c.reconnectBackoff = initialReconnectBackoff
 
 	c.debug("ws: connected successfully")
 
@@ -314,6 +332,39 @@ func (c *WSClient) handleDisconnect(conn *websocket.Conn) {
 	go c.reconnect()
 }
 
+// nextReconnectDelay returns how long the caller should wait before its next
+// reconnect attempt, and advances the schedule for the attempt after that.
+//
+// The read, the doubling and the clamp all happen in ONE critical section under
+// connMu, and the caller sleeps on the returned value instead of re-reading the
+// field. reconnect used to do all of it with no lock at all, racing the reset in
+// connectLocked, which runs under connMu (#728). It also read the field twice,
+// once to log the delay and once to sleep it, so a reset landing between the two
+// made the node retry sooner than it had just announced.
+//
+// Both defects point the same way, and it is the way that hurts: anything that
+// silently shortens a backoff moves toward the #443 restart storm referenced
+// above connectWithBackoff in cmd/work.go, where tight retries burned the node's
+// daily Redis-API quota and locked it out for about a day.
+//
+// connMu is reused deliberately rather than adding a third mutex to this file.
+// The reset has to live in connectLocked, which is the only place that knows a
+// dial succeeded and whose caller already holds connMu, so connMu costs no extra
+// lock and no nested acquire. connMu is NOT held across the sleep.
+func (c *WSClient) nextReconnectDelay() time.Duration {
+	c.connMu.Lock()
+	defer c.connMu.Unlock()
+
+	delay := c.reconnectBackoff
+
+	c.reconnectBackoff *= 2
+	if c.reconnectBackoff > c.maxBackoff {
+		c.reconnectBackoff = c.maxBackoff
+	}
+
+	return delay
+}
+
 // reconnect attempts to reconnect with exponential backoff
 func (c *WSClient) reconnect() {
 	for {
@@ -331,14 +382,9 @@ func (c *WSClient) reconnect() {
 			return
 		}
 
-		c.debug("ws: attempting reconnect in %v", c.reconnectBackoff)
-		time.Sleep(c.reconnectBackoff)
-
-		// Increase backoff for next attempt
-		c.reconnectBackoff *= 2
-		if c.reconnectBackoff > c.maxBackoff {
-			c.reconnectBackoff = c.maxBackoff
-		}
+		delay := c.nextReconnectDelay()
+		c.debug("ws: attempting reconnect in %v", delay)
+		time.Sleep(delay)
 
 		c.connMu.Lock()
 		err := c.connectLocked(context.Background())
--- a/internal/redisapi/websocket.go
+++ b/internal/redisapi/websocket.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/rand/v2"
 	"net"
 	"net/http"
 	"net/url"
@@ -91,6 +92,13 @@ type WSClient struct {
 	reconnectBackoff time.Duration
 	maxBackoff       time.Duration
 
+	// jitterFrac draws the randomness nextReconnectDelay adds on top of the
+	// schedule. It is a field rather than a package var so a test can make the
+	// delay deterministic without mutating shared state that a previous test's
+	// still-running reconnect goroutine might be reading. Set once in
+	// NewWSClient, before any goroutine exists. Must return [0,1).
+	jitterFrac func() float64
+
 	// Reconnect callbacks fired after a successful reconnection
 	reconnectCallbacks   []func()
 	reconnectCallbacksMu sync.RWMutex
@@ -139,6 +147,56 @@ type WSMessage struct {
 // one place so the constructor and the reset in connectLocked cannot drift.
 const initialReconnectBackoff = time.Second
 
+// Reconnect jitter.
+//
+// The schedule used to be exactly 1, 2, 4, ... 60 seconds on every node, with no
+// randomness anywhere in this file. Whenever the control plane blips, every node
+// in the fleet is disconnected at nearly the same instant and then redials in
+// lockstep, against an endpoint that answers 429 (typed by rateLimitFromUpgrade
+// above). That is a self-inflicted thundering herd, and the keepalive in #734
+// makes it worse rather than better: handleDisconnect now fires on every
+// read-deadline expiry instead of almost never, so this path runs routinely.
+//
+// The scheme is ADDITIVE and UPWARD-ONLY: the sleep is drawn uniformly from
+// [base, 2*base), where base is the deterministic step. The two textbook
+// alternatives were both rejected for the same reason, which is the only
+// property this package cannot trade away:
+//
+//   - Full jitter, rand(0, base), has a floor of zero. A node that has just been
+//     told 429 could redial milliseconds later, and the mean delay is HALF the
+//     current schedule, so fleet-wide retry pressure doubles. That is the shape
+//     of #443, where tight retries burned the node's daily Redis-API quota and
+//     locked it out for about a day, and it is the direction #728 calls out as
+//     the one that matters. cmd/work.go's rate-limit path already refuses to
+//     "poll tighter than retry_after" for the same reason.
+//   - Decorrelated jitter, min(cap, rand(base, prev*3)), can also return to base
+//     repeatedly, so a long outage can sit near one second indefinitely. It also
+//     wants the previous SLEEP as its state, which would collide with
+//     reconnectBackoff holding the deterministic schedule that connectLocked
+//     resets on a successful dial.
+//
+// Additive jitter is never shorter than what shipped before, and its spread is
+// the full width of the current step, which is exactly enough to decorrelate a
+// fleet whose nodes all hold the same step after a shared blip.
+//
+// The jittered sleep is deliberately NOT clamped to maxBackoff. maxBackoff caps
+// the SCHEDULE, not the sleep. Clamping the result would collapse the spread to
+// zero at the ceiling, which is precisely the worst case: a long control-plane
+// outage puts every node at the cap, and they would all dial together again.
+// The cost is a worst-case sleep of two minutes rather than one, reached only
+// after roughly seven consecutive failed dials, and publishes fall back to HTTP
+// throughout (see Client.Publish).
+//
+// Note what this does NOT fix, so nobody reads more into it than is there: the
+// flap in #734's own "what this makes worse" (a peer that is alive but never
+// pongs, torn down every pongWait, reconnecting at one second, forever) has a
+// period dominated by pongWait, not by the backoff. One second of jitter per
+// 46-second cycle random-walks those phases apart far too slowly to matter. Two
+// things would actually address it, both filed rather than done here: not
+// resetting the schedule for a connection that never proved itself, and honoring
+// the typed rate-limit hint that connectLocked already builds and reconnect
+// currently discards.
+
 // NewWSClient creates a new WebSocket client.
 func NewWSClient(cfg WSClientConfig) *WSClient {
 	reconnectEnabled := cfg.ReconnectEnabled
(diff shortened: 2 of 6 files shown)


sunapi386:
CI green on `integrate/728-734-jitter` (run 31458391339, success). Worth knowing before merging: the `-race` run of `internal/redisapi` went from roughly 21s to roughly 29s locally, because the combined regression test rides out about twenty read-deadline cycles and the #741 test waits five times its backoff before concluding no dial happened. Both budgets are deliberately slack rather than tight, for the reason `backoffSleep` in `cmd/work.go` documents: on a contended runner a tight budget fails in the direction that gets a build rerun instead of diagnosed.

Issue #739, opened 2026-08-11
Full-wipe guard only covers the empty desired set: one desired module would uninstall every other manifest service


sunapi386:
## What is wrong

`Reconciler.RefuseFullWipe` only guards the **empty** desired set. A desired set
with one module still converges authoritatively, and
`liveModuleOps.ListInstalled` (`cmd/module_ops.go`) enumerates every entry in
`citadel.yaml` `services:`, not just modules the control plane installed.

So the first non empty desired state assigned to a node uninstalls every
manifest service that is not in it. On a node whose manifest carries 11 services
(vllm, ollama, llamacpp, transcribe, gotenberg, meeting, kokoro, tei-embeddings,
unlimited-ocr, bonsai, tei), a single desired row means 10 compose stacks are
torn down, and `liveModuleOps.Uninstall` also drops them from the manifest,
deletes their lockfile entries, and removes their materialized compose and env
files.

This is latent today only because nothing writes the durable desired rows yet.
The aceteam side records the same hazard in its module tool docstring: durable
desired state persistence "remains a follow-up so a lone desired row cannot make
the eventual pull loop uninstall every other module." The node side should not
depend on the control plane never doing that.

## Why now

Found while fixing #733. That fix makes a refused pass report its observed set,
which surfaces all 11 manifest services as reported "modules". Anyone reading
that list could reasonably assign desired state from it.

## Options

1. Scope the reconcile engine to modules the module system actually installed
   (lockfile entries) rather than every manifest service, so plain services and
   embedded engines are outside its authority.
2. Keep enumerating everything but refuse any pass whose plan would uninstall
   more than N modules, or more than it installs, as a proportional version of
   the full-wipe guard.
3. An explicit per node opt in before the pull loop is allowed to uninstall
   anything at all.

Option 1 looks closest to the intended model: the reconcile package's own doc
describes modules as the unit, and a service started by `citadel run` or an
operator was never module managed.

sunapi386:
[...]

PR #752, opened 2026-08-13
reconcile: never-managed nodes should not trip the full-wipe guard forever (#733)

sunapi386:
## Context

Node 1297 (and effectively every fabric node today) trips
`internal/reconcile/loop.go`'s full-wipe guard every single reconcile pass,
forever: `reconcile: refusing empty desired state with N module(s) installed
(possible control-plane misconfiguration)`. PR #737 already fixed the
observability half (a refused pass still reports its observed actual state).
This PR is the deeper defect that #737 exposed but didn't fix: the guard
itself cannot tell "this node was never assigned anything" apart from "this
node was assigned modules and the control plane is now serving zero", and it
treats both as an error, forever.

## Root cause

Traced the full data flow end to end:

- `internal/reconcile/loop.go` fetches desired state via `DesiredStateProvider`
  then `ProtoProvider` then `redisapi.Client.GetDesiredState`, which calls
  `GET /api/fabric/nodes/{id}/desired-state`.
- On the aceteam side, `python-backend/routes/fabric_desired_state.py` serves
  that endpoint by reading the `fabric_node_module_desired` table.
- **Nothing writes that table.** Confirmed by grep across the whole aceteam
  repo: only two files even reference `fabric_node_module_desired` (the read
  endpoint and a docstring). The intended writer, `fabric_node_module_set`
  (the MCP tool for setting a node's desired module state), explicitly does
  NOT persist, by its own docstring: *"it does NOT persist the desired
  state server-side... Durable desired-state storage
  (`fabric_node_module_desired`)... remains a follow-up (aceteam #4273)"*.
- Confirmed empirically against the live DB: the table exists (schema already
  matches what the read endpoint expects) and has exactly 5 rows total, all
  for one other node, all inserted at the same timestamp (a manual seed, not
  a live writer). Zero rows for node 1297.
- So `desired.Modules` is permanently `[]` for every real node, `RefuseFullWipe`
  fires on every pass that has any actual module installed, and it logs/errors
  forever with no way to converge.

## Why the fix is scoped the way it is (and why the "obvious" fix is unsafe)

The obvious-looking fix, having `fabric_node_module_set` actually write to
`fabric_node_module_desired`, is a genuine architecture fork, not a
mechanical bug, and I did **not** implement it here. Filed separately (and
already open, unassigned) as **#739**: the reconcile engine is fully
declarative once `desired.Modules` is non-empty, and `liveModuleOps.
ListInstalled` enumerates every `citadel.yaml` service, not just modules the
module system installed. So the FIRST desired row ever written for a
never-managed node with N other manifest services running (vllm, ollama,
transcribe, gotenberg, bonsai, ...) would uninstall the other N-1 the very
next pass, turning a merely-noisy bug into a genuinely destructive one. #739
lays out three real options (scope reconcile authority to lockfile-tracked
modules only / cap the uninstall blast radius / require explicit node opt-in)
and none of them are mine to pick.

**What I built instead** is safe regardless of how #739 resolves, because it
never touches the write path or the diff/apply logic at all:

`DesiredState.NeverManaged()` (`internal/reconcile/types.go`) reports whether
the control plane has ever assigned this node anything: `Revision == ""` or
`"0"`, which the wire-level revision handshake already carries and which the
control plane serves only when it holds no `fabric_node_module_desired` row
for this node at all (paired aceteam-side change makes this true even across
a soft-deleted history, see that PR).

`ReconcileOnce` (`internal/reconcile/loop.go`) now splits the "empty desired
with modules installed" case in two:
- **Never managed** (`NeverManaged() == true`): nothing was ever assigned, so
  there is nothing to refuse. Skip `Reconcile`/`Apply` (an empty, successful
  plan, identical safety to today: nothing is ever uninstalled) and fall
  through to the normal report tail, revision handshake included. No error,
  no per-pass log noise, and the node's actual state gets reported like any
  other converged pass.
- **Actually managed and now empty** (`NeverManaged() == false`): unchanged,
  refuse, error, report-without-revision-stamp, exactly as PR #737 shipped it.

Neither branch ever calls `Reconcile`/`Apply`, so this cannot produce an
uninstall plan in either case; it only changes whether the (still-refused)
pass is treated as an error.

## Companion aceteam-side change

A small paired PR in `aceteam-ai/aceteam` makes `revision` survive a
soft-deleted history (queries the row watermark ignoring `delete_on`) so a
node that WAS managed and later had everything explicitly removed still
reports a non-zero revision, keeping the managed/never-managed distinction
correct even after #739 unblocks real writes. Not required for today's bug
(nothing writes/deletes yet), included as defense-in-depth so it isn't
forgotten later. Linked in that PR's description.

## Schema changes

None. `fabric_node_module_desired` already exists with the exact shape the
read endpoint uses; this PR does not write to it.

## Architecture fork for Jason (not decided here)

**citadel#739** (already filed, unassigned): once desired-state persistence
is added, the reconcile engine treats "the control plane assigned zero
modules" as authoritative and uninstalls everything else on the node's
manifest that isn't in that set. Options laid out there:
1. Scope reconcile authority to modules the module system actually installed
   (lockfile entries), not every manifest service.
2. Keep full enumeration, but cap any single pass's uninstall blast radius.
3. Require an explicit per-node opt-in before the pull loop may uninstall
   anything.

My read: option 1 fits the package's own stated model best (`internal/
reconcile`'s doc already describes modules, not "everything in citadel.yaml",
as the unit), but it's a real product/semantics call, not mine to make.

## Testing

`go test ./...`: full suite green, no regressions.

New tests in `internal/reconcile/fullwipe_test.go`:
- `TestDesiredStateNeverManaged`: pins the predicate (`""`/`"0"` -> true,
  anything else, including an opaque non-numeric revision token, -> false).
- `TestRefuseFullWipeSkipsNeverManagedNode` / `...EmptyRevision`: a
  never-managed node with modules installed converges as a no-op: no error,
  no uninstall calls, and the observed actual state IS still reported (with
  the revision handshake stamped, unlike the still-refused/errored path).

New test in `internal/reconcile/proto_provider_test.go`:
- `TestNeverManagedThroughRealProtoWireNoUninstall`: the tests above construct
  the internal `DesiredState` directly, which never exercises the real wire
  path (marshal, HTTP body, unmarshal, `adaptDesiredState`). This one drives a
  real `fabricpb.DesiredState{Revision: "0"}` through `ProtoProvider.Fetch`
  into a `Reconciler` with two already-installed modules, asserting no error,
  no uninstall calls, and the actual state still reported.

Mutation-tested: reverted `loop.go`/`types.go` to `origin/main` (via
`git checkout origin/main -- <files>`, not stash: this repo's shared-stash
gotcha), confirmed the new tests fail (compile error, since `NeverManaged`
no longer exists), restored via a saved patch, reran; green again. Existing
tests (`TestRefuseFullWipeBlocksEmptyDesiredWithInstalled` and friends, which
use non-zero opaque revisions like `"rev-empty"`) were unaffected, confirming
the managed/error path is unchanged. Separately mutation-tested
`TestNeverManagedThroughRealProtoWireNoUninstall` by corrupting
`adaptDesiredState`'s revision copy to a nonzero sentinel (`Revision:
"corrupted-nonzero"`): the test failed with the guard error, confirming it
actually exercises the proto decode path; restored and reran green.

## Draft

Leaving this in draft: it touches reconcile/dispatch control flow and should
get a look before merge, per this repo's CLAUDE.md.

### DIFF
--- a/internal/reconcile/fullwipe_test.go
+++ b/internal/reconcile/fullwipe_test.go
@@ -129,6 +129,92 @@ func TestRefuseFullWipeSurfacesReportFailure(t *testing.T) {
 	}
 }
 
+// TestRefuseFullWipeSkipsNeverManagedNode is the #733 root-cause regression: a
+// node the control plane has NEVER assigned any desired state to (Revision ==
+// "0", the zero value the control plane serves when it holds no
+// fabric_node_module_desired row for this node at all) must not be treated as
+// "misconfigured". Before this fix EVERY node hit this path forever, because
+// nothing writes that durable store yet — the guard fired and errored on every
+// single pass (4574+ times over 5+ days, observed on node 1297). It must now
+// succeed as a no-op converge (nothing installed, nothing uninstalled) while
+// still reporting the observed actual state, exactly like a legitimate
+// converged pass.
+func TestRefuseFullWipeSkipsNeverManagedNode(t *testing.T) {
+	ops := newFakeOps(
+		InstalledModule{Name: "a", Source: "src-a", Health: HealthRunning},
+		InstalledModule{Name: "b", Source: "src-b", Health: HealthRunning},
+	)
+	provider := &FakeProvider{Desired: DesiredState{Revision: "0"}} // never managed
+	rec := NewReconciler(provider, ops, "node-1297")
+	rec.RefuseFullWipe = true
+
+	plan, _, err := rec.ReconcileOnce(context.Background())
+	if err != nil {
+		t.Fatalf("never-managed node with modules installed must succeed as a no-op, got: %v", err)
+	}
+	if !plan.IsEmpty() {
+		t.Errorf("want empty plan (nothing to converge to), got %+v", plan.Steps)
+	}
+	for _, c := range ops.calls {
+		if strings.HasPrefix(c, "uninstall:") {
+			t.Fatalf("a never-managed node must not uninstall anything, saw %q", c)
+		}
+	}
+	if len(provider.Reported) != 1 {
+		t.Fatalf("want exactly 1 report, got %d", len(provider.Reported))
+	}
+	report := provider.Reported[0]
+	if len(report.Modules) != 2 {
+		t.Fatalf("want 2 reported modules, got %d (%+v)", len(report.Modules), report.Modules)
+	}
+	// Unlike the refused-and-errored path, this IS a (no-op) converge, so the
+	// revision handshake stamps normally.
+	if report.AppliedRevision != "0" {
+		t.Errorf("AppliedRevision = %q, want %q (echoed like any other converge)", report.AppliedRevision, "0")
+	}
+}
+
+// TestRefuseFullWipeSkipsNeverManagedNodeEmptyRevision confirms the same
+// no-op behavior when Revision is the empty string, not just the literal "0"
+// — both are DesiredState.NeverManaged().
+func TestRefuseFullWipeSkipsNeverManagedNodeEmptyRevision(t *testing.T) {
+	ops := newFakeOps(InstalledModule{Name: "a", Source: "src-a", Health: HealthRunning})
+	provider := &FakeProvider{Desired: DesiredState{Revision: ""}}
+	rec := NewReconciler(provider, ops, "node")
+	rec.RefuseFullWipe = true
+
+	plan, _, err := rec.ReconcileOnce(context.Background())
+	if err != nil {
+		t.Fatalf("never-managed node (empty revision) must succeed as a no-op, got: %v", err)
+	}
+	if !plan.IsEmpty() {
+		t.Errorf("want empty plan, got %+v", plan.Steps)
+	}
+}
+
+// TestDesiredStateNeverManaged pins the predicate directly: only the true
+// zero-history values ("" and "0") count as never-managed; any other revision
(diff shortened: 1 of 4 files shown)


sunapi386:
Companion aceteam-side PR (private repo): aceteam-ai/aceteam#7526 (keeps the desired-state revision watermark correct across a soft-deleted history).

sunapi386:
Pushed a follow-up commit:

- Added TestNeverManagedThroughRealProtoWireNoUninstall: the existing NeverManaged tests construct the internal DesiredState directly (FakeProvider), which never exercises adaptDesiredState's proto to internal mapping. This one marshals a real fabricpb.DesiredState with Revision "0" and no modules through ProtoProvider.Fetch into a Reconciler with two already-installed modules, asserting no error, no uninstall calls, and the actual state still reported. Mutation-tested by corrupting adaptDesiredState's revision copy to a nonzero sentinel (confirmed the test fails), then to a dropped/empty copy (confirmed the two existing revision-carry tests already catch that case), then restored.
- Removed em dashes that slipped into the comments added for this fix in loop.go, types.go, and fullwipe_test.go.

Confirmed adaptDesiredState does copy Revision through verbatim (`DesiredState{Revision: pb.GetRevision()}`), so the wire path is safe; this test now pins that permanently instead of by inspection.

go test ./... still green.

